PR #2, opened 2016-09-23
Remove extra unknown props

styks1987:
https://facebook.github.io/react/warnings/unknown-prop.html

### DIFF
--- a/lib/number_format.js
+++ b/lib/number_format.js
@@ -152,19 +152,20 @@ var NumberFormat = React.createClass({
   render: function render() {
     var props = _extends({}, props);
     delete props.thousandSeparator;
-    delete props.displayType;
     delete props.suffix;
     delete props.prefix;
     delete props.format;
     delete props.mask;
 
     if (props.displayType === "text") {
+      delete props.displayType;
       return React.createElement(
         'span',
         props,
         this.state.value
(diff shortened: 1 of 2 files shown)


lib/number_format.js
```diff
@@ -150,16 +150,23 @@ var NumberFormat = React.createClass({
     this.onChangeHandler(e, this.props.onInput);
   },
   render: function render() {
-    var props = this.props;
```
s-yadav: Can you make this changes on src files instead of lib?

lib/number_format.js
```diff
@@ -150,16 +150,23 @@ var NumberFormat = React.createClass({
...
+    var props = _extends({}, {}, this.props);
```
s-yadav: Why double objects. It can be just 
```js
var props = _extends({},  this.props);
```

Also we can minimze by code by 
```js
['thousandSeperator', 'suffix', 'prefix', /* 'all others custom props' */].forEach((key) => {
 delete props[key];
});
```

styks1987: Updated with changes

s-yadav:
For future fixes can you follow the development guide mentioned in readme. Let me know if you find any difficulty setting it up locally.

PR #6, opened 2016-11-03
Handle onBlur event

saidmoya12:
Handle onBlur event

### DIFF
--- a/src/number_format.js
+++ b/src/number_format.js
@@ -123,8 +123,15 @@ const NumberFormat =  React.createClass({
 
     return value;
   },
+  onBlurHandler: function(e, callback){
+    const {value} = this.state;
+    if(callback) callback(e,value);
+  },
   onChange : function(e){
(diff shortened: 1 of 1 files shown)


src/number_format.js
```diff
@@ -123,9 +123,16 @@ const NumberFormat =  React.createClass({
...
     return value;
   },
+  onBlurHandler: function(e, callback){
+    const {value} = this.state;
+    if(callback) callback(e,value);
+  },
   onChange : function(e){
     this.onChangeHandler(e,this.props.onChange);
   },
+  onBlur : function(e){
```
s-yadav: I understand that you need the non-formatted value in blur event
You keep your state sync using onChange event, and then use it on onBlur event
```js
< NumberFormat onChange = {(e, value) => {this.setState({card: value})}} onBlur={() => {
  const {card} = this.state;
}} {...otherOptions} />
```

PR #9, opened 2016-11-22
Persist event in onChange handler

mikker:
Accessing the Event asynchronously doesn't work since React 0.14. This fixes this Warning:

```
Warning: This synthetic event is reused for performance reasons. If you're seeing this, you're accessing the property `target` on a released/nullified synthetic event. This is set to null. If you must keep the original synthetic event around, use event.persist(). See https://fb.me/react-event-pooling for more information.
```


s-yadav:
Applied those changes on latest version 1.0.0 . Thanks

PR #13, opened 2016-12-13
Make input type variable

lgraubner:
Currently the input type is hardcoded to `tel`. That doesn't suit all use cases. Therefore I made `type` a property, defaulting to `text`. Currently only `tel` and `text` are allowed. E.g. `number` does not work. There might be more usable types I didn't tested. Feel free to add them in the `propTypes` object.

Also I added the missing dep `cross-env` to make the tests work out of the box.


s-yadav:
Type tel was set so mobile brings the number pad. For desktop it doesn't matter. What are the use cases in which tel doesn't work ?

lgraubner:
`tel` works fine. I just like to have semantic correct input types. Mobile is not that important in my case, hence the pull request.
So `tel` as default seems to make more sense.

davidmillington:
Hi @s-yadav, looks like this would help with issue #35 and still be backwards compatible.
Rather than just me adding a +1 on the issue, is there anything I help with in order to get this pull request through?

kentandersen:
Merged this pull request with master (scaling back the syntactical changes), and created a new pull request at #57

s-yadav:
Added on 2.0.0

PR #15, opened 2016-12-15
decimalPrecision option and supporting tests

styks1987:
This allows the user to pass a decimalPrecision option and the number will format based on the precision. I know this could be done with the format but it is nice to have working in conjunction with prefix and suffixes.

### DIFF
--- a/src/number_format.js
+++ b/src/number_format.js
@@ -8,6 +8,7 @@ function escapeRegExp(str) {
 const propTypes = {
   thousandSeparator: PropTypes.oneOf([',', '.', true, false]),
   decimalSeparator: PropTypes.oneOf([',', '.', true, false]),
+  decimalPrecision: PropTypes.oneOfType([PropTypes.number, PropTypes.bool]),
   displayType: PropTypes.oneOf(['input', 'text']),
   prefix: PropTypes.string,
   suffix: PropTypes.string,
@@ -24,7 +25,8 @@ const propTypes = {
 
 const defaultProps = {
   displayType: 'input',
-  decimalSeparator: '.'
+  decimalSeparator: '.',
+  decimalPrecision: false
 };
 
 class NumberFormat extends React.Component {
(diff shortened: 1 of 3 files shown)


src/number_format.js
```diff
@@ -24,7 +25,8 @@ const propTypes = {
 
 const defaultProps = {
   displayType: 'input',
-  decimalSeparator: '.'
+  decimalSeparator: '.',
+  decimalPrecision: false
```
s-yadav: Keep then default value of decimal precision as unlimited

src/number_format.js
```diff
@@ -139,9 +142,14 @@ class NumberFormat extends React.Component {
     }
     else{
       let beforeDecimal = formattedValue, afterDecimal = '';
-      const hasDecimals = formattedValue.indexOf(decimalSeparator) !== -1;
+      const hasDecimals = formattedValue.indexOf(decimalSeparator) !== -1 || decimalPrecision !== false;
```
s-yadav: Based on current default value of decimalPrecision, user always have to define decimalPrecision otherwise he will not be able to add it.

src/number_format.js
styks1987: Not sure I understand your comment. If they pass false to the decimalPrecision, it would be unlimited number of decimal places. Could you explain this a little more?

src/number_format.js
styks1987: Not be able to add what? If they don't define decimalPrecision, it would bypass my change altogether and function as it normally would. Am I missing something?

bengtrj:
I like the library but the lack of a way to have the decimals always shown is keeping me from using it. The patch is perfect and includes tests, I hope is incorporated soon!!

s-yadav:
@styks1987 Ignore the previous comments. Code looks good. Though there are some fixes required.
1. Eslint error
2. If we press backspace just after decimal it corrupt the number, ie $23,498.00 becomes $23,49,800.00

I will see what I can do fo this in this weekend.

PR #38, opened 2017-03-08
negative number initial commit

joe-vh:
Let me know if this is of use or any extra behaviour is needed and I will update.

### DIFF
--- a/src/number_format.js
+++ b/src/number_format.js
@@ -21,20 +21,23 @@ const propTypes = {
     PropTypes.number,
     PropTypes.string
   ]),
+  allowNegative: PropTypes.bool,
   customInput: PropTypes.func
 };
 
 const defaultProps = {
   displayType: 'input',
   decimalSeparator: '.',
-  decimalPrecision: false
+  decimalPrecision: false,
+  allowNegative: true
 };
 
 class NumberFormat extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      value: this.formatInput(props.value).formattedValue
+      value: this.formatInput(props.value).formattedValue,
+      negative: false
     }
     this.onChange = this.onChange.bind(this);
     this.onKeyDown = this.onKeyDown.bind(this);
@@ -43,7 +46,8 @@ class NumberFormat extends React.Component {
   componentWillReceiveProps(newProps) {
     if(newProps.value !== this.props.value) {
       this.setState({
-        value : this.formatInput(newProps.value).formattedValue
+        value : this.formatInput(newProps.value).formattedValue,
+        negative: !newProps.value.length ? false : newProps.allowNegative && this.state.negative
       });
     }
   }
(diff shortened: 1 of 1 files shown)


src/number_format.js
```diff
@@ -214,10 +219,13 @@ class NumberFormat extends React.Component {
   onKeyDown(e) {
     const el = e.target;
     const {selectionStart, selectionEnd, value} = el;
-    const {decimalPrecision} = this.props;
+    const {decimalPrecision, allowNegative} = this.props;
     const {key} = e;
     const numRegex = this.getNumberRegex(false, decimalPrecision !== false);
     //Handle backspace and delete against non numerical/decimal characters
+    if (key === '-' && allowNegative) {
+      this.setState({negative: !this.state.negative});
```
alexandrapersea: This way if you delete the value from the input and type another one, it will still be negative. Maybe you can find a workaround for that.

src/number_format.js
```diff
@@ -21,20 +21,23 @@ const propTypes = {
...
+  allowNegative: PropTypes.bool,
   customInput: PropTypes.func
 };
 
 const defaultProps = {
   displayType: 'input',
   decimalSeparator: '.',
-  decimalPrecision: false
+  decimalPrecision: false,
+  allowNegative: true
```
alexandrapersea: Why is allowNegative true as default? Shouldn't it be the other way around?

alexandrapersea:
This is a good feature 💯 . Could you write some test for it?

src/number_format.js
joe-vh: I deal with negative liquidity every day so it seemed like a reasonable default financial use cases, this can be changed to false

src/number_format.js
joe-vh: Yes I will try to find time to do this - in my workplace I have a version that does this but it it integrated into an old version of this library.

s-yadav: Also, this will act as toggle, you will be able to press - at any cursor position and that will toggle the -, which shouldn't happen, user should be only allowed to type it before the number.

src/number_format.js
```diff
@@ -161,6 +165,7 @@ class NumberFormat extends React.Component {
       if(thousandSeparator) {
         beforeDecimal = beforeDecimal.replace(/(\d)(?=(\d{3})+(?!\d))/g, '$1' + thousandSeparator);
       }
+      beforeDecimal = allowNegative && this.state && this.state.negative ? '-' + beforeDecimal : beforeDecimal;
```
s-yadav: negation sign should come before prefix, as the general format is 
`-$13,344.23`

src/number_format.js
s-yadav: I think default value should be true if format prop is not defined and should be false if format props defined.

s-yadav:
Works correctly on 2.0.0. This pr is no longer required

PR #51, opened 2017-04-25
Migrate deprecated PropTypes

slate71:
This add the `prop-types` package and migrates the import from React to the prop-types package.


sidmitra:
@s-yadav I think this should be merged, to match how PropTypes are used now. I tested it and works fine.

ranneyd:
+1

ranneyd:
@s-yadav merge pls

PR #55, opened 2017-05-05
fix: re-render on (e.g. prefix prop) update

a-x-:
When prefix (or any another prop) is changing react-number-format must rerender itself.

Also, you can Look for `PureComponent` or `shouldComponentUpdate` for better render optimization technics.


s-yadav:
Fixed on 2.0.0

PR #59, opened 2017-05-11
Fix race conditions by setting cursor position synchronously

stutrek:
Entering text quickly with an onChange that takes a long time causes a race condition. This fixes that by moving the evaluation and setting of the cursor position from the async callback from state change to the callback of the input itself. It also adds a clearTimeout for the async movement needed for Chrome Mobile.

With a long running onChange and by typing quickly, I am consistently able to get events in this order:

1. Enter first character 
2. Get state change callback from first, call long running onChange callback
3. Enter second character
4. Enter third character
5. Get state change callback from second character, move cursor to position 2 because that's how long the value from the second event is
6. Get state change callback from third character, detect cursor at position 2 and leave it there

Here's what I'm seeing in my console when I add a log:
screen shot 2017-05-11 at 4 38 42 pm

It lists 2 twice, then skips 3 and goes to 4 because it added a comma.

Putting timeout directly on `this` instead of `this.state` is needed in this situation because it must be immediately available and because changing its value should not cause a rerender.

### DIFF
--- a/src/number_format.js
+++ b/src/number_format.js
@@ -231,17 +231,29 @@ class NumberFormat extends React.Component {
     const inputValue = el.value;
     const {formattedValue,value} = this.formatInput(inputValue);
     let cursorPos = el.selectionStart;
+    
+    cursorPos = this.getCursorPosition(inputValue, formattedValue, cursorPos );    
+    
+    el.value = formattedValue;
+    this.setCaretPosition(el, cursorPos);
+
+    /*
+      setting caret position within timeout of 0ms is required for mobile chrome,
+      otherwise browser resets the caret position after we set it
+      We are also setting it without timeout so that in normal browser we don't see the flickering
+     */
+    if (this.timeout) {
+      clearTimeout(this.timeout);
+    }
+    this.timeout = setTimeout(() => {
+      this.timeout = null;
+      if (el.value === formattedValue) {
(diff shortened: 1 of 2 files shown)


wrabbit23:
This looks like the fix I need

s-yadav:
Caret position need to be set after setState as formattedValue can be different from input value.

stutrek:
This takes that into account by setting the formattedValue before calling setState.

src/number_format.js
```diff
@@ -231,17 +231,29 @@ class NumberFormat extends React.Component {
...
+    el.value = formattedValue;
+    this.setCaretPosition(el, cursorPos);
+
+    /*
+      setting caret position within timeout of 0ms is required for mobile chrome,
+      otherwise browser resets the caret position after we set it
+      We are also setting it without timeout so that in normal browser we don't see the flickering
+     */
+    if (this.timeout) {
+      clearTimeout(this.timeout);
```
s-yadav: Is this required because I guess onChangeHandler will be called on next event cycle before which caret position timeout with 0ms will already be executed

stutrek: It would happen if the computer is busy and two key presses get put into the event queue before the timeout.

s-yadav:
Fixed on 2.0.0

PR #60, opened 2017-05-14
Add limit max feature plus matching tests

shirilad:
With max property given, the code will now verify that the given value is indeed smaller than max. 
If not, it won't allow the value, and return the value without its most right digit.


shirilad:
The generated files merge might be a bit off. Hope I didn't mess something up there.
I'm guessing a clean bundle run will solve it anyways, but if there's something there needs fixing lmk and I'll try to solve accordingly.

s-yadav:
Instead of max prop added isAllowed prop to reject a value

PR #76, opened 2017-07-31
Allow empty string value when decimalPrecision is passed

### LINKED ISSUE #75: Not possible to render empty string if decimalPrecision is passed
I mentioned this in issue #70, but if you pass a value of `''` and set the decimalPrecision prop, the input will render `0.00` instead of nothing. This makes it impossible to use the `placeholder` attribute on your input, and is inconsistent with the behaviour when you *don't* pass `decimalPrecision` and it is possible to render nothing.

adamshone:
Fixes #75


s-yadav:
Looks good. I will test and merge.

PR #80, opened 2017-08-19
Fix linter warnings. Fixes #77

### LINKED ISSUE #77: Fails if value = null
Component fails on null value in props

isnifer:
Hi! Thank you for the component. I want to help with #77

### DIFF
--- a/src/number_format.js
+++ b/src/number_format.js
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 
-function noop(){};
+function noop(){}
 
 function escapeRegExp(str) {
   return str.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&");
@@ -84,6 +84,7 @@ const defaultProps = {
   suffix: '',
   allowNegative: true,
   type: 'text',
+  value: '',
   onChange: noop,
   onKeyDown: noop,
   onMouseUp: noop,
@@ -128,7 +129,7 @@ class NumberFormat extends React.Component {
   getFloatString(num, props) {
     props = props || this.props;
     const {decimalSeparator, thousandSeparator} = this.getSeparators(props);
-    return num.replace(new RegExp(escapeRegExp(thousandSeparator || ''), 'g'), '').replace(decimalSeparator, '.');
+    return (num || '').replace(new RegExp(escapeRegExp(thousandSeparator || ''), 'g'), '').replace(decimalSeparator, '.');
   }
 
   getFloatValue(num, props) {
@@ -142,7 +143,7 @@ class NumberFormat extends React.Component {
 
     let {value} = props;
 
-    if (format || value === undefined || value === '') return value;
+    if (format || !(value || value === 0)) return value;
 
     const isNumber = typeof value === 'number';
 
@@ -407,7 +408,7 @@ class NumberFormat extends React.Component {
     const {state, props} = this;
     const {isAllowed} = props;
     const lastValue = state.value;
-    let {formattedValue, value} = this.formatInput(inputValue);
+    let {formattedValue, value} = this.formatInput(inputValue); // eslint-disable-line prefer-const
 
     /*Max of selectionStart and selectionEnd is taken for the patch of pixel and other mobile device caret bug*/
     const currentCaretPosition = Math.max(el.selectionStart, el.selectionEnd);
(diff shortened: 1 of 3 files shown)


src/number_format.js
```diff
@@ -128,12 +129,14 @@ class NumberFormat extends React.Component {
   getFloatString(num, props) {
     props = props || this.props;
     const {decimalSeparator, thousandSeparator} = this.getSeparators(props);
-    return num.replace(new RegExp(escapeRegExp(thousandSeparator || ''), 'g'), '').replace(decimalSeparator, '.');
+    const number = num || '';
```
s-yadav: This can be moved to arguments 
```js
getFloatString(num='', props)
```

src/number_format.js
```diff
@@ -128,12 +129,14 @@ class NumberFormat extends React.Component {
...
+    return number.replace(new RegExp(escapeRegExp(thousandSeparator || ''), 'g'), '').replace(decimalSeparator, '.');
   }
 
   getFloatValue(num, props) {
     props = props || this.props;
-    return parseFloat(this.getFloatString(num, props)) || 0;
+    const number = num || '';
```
s-yadav: Move to argument

src/number_format.js
```diff
@@ -407,7 +410,8 @@ class NumberFormat extends React.Component {
     const {state, props} = this;
     const {isAllowed} = props;
     const lastValue = state.value;
-    let {formattedValue, value} = this.formatInput(inputValue);
+    const {value} = this.formatInput(inputValue);
```
s-yadav: Calling formatInput twice is not a good idea just to maintain const and let.
You can probably store it in some different variable. 
```js
const values = this.formatInput(inputValue);
const {value} = values;
let {formattedValue} = values;
```
Or I will prefer just to disable eslint for this line
  ```js 
 let {formattedValue, value} = this.formatInput(inputValue); // eslint-disable-line prefer-const
```

File: src/number_format.js (hunk 1 above)
Comment (isnifer) [inline src/number_format.js]: No no, it will not override `null`. Only if you pass `undefined` as the first argument, num will be `''`

File: src/number_format.js (hunk 2 above)
Comment (isnifer) [inline src/number_format.js]: the same

**Comment by isnifer** [comment]
@s-yadav hey

File: src/number_format.js (hunk 1 above)
Comment (s-yadav) [reply src/number_format.js]: Ohh yes, 
Instead of creating new const just change it to
```js
num = num || '';
```
or
```js
return (num || '').replace(new RegExp(escapeRegExp(thousandSeparator || ''), 'g'), '').replace(decimalSeparator, '.');
```

src/number_format.js
s-yadav: same

s-yadav:
Also can you make a check for null here as well
https://github.com/isnifer/react-number-format/blob/028d767201f2d227e53719ed0af8ffe5d9dc7e5d/src/number_format.js#L148

isnifer:
@s-yadav got it https://github.com/s-yadav/react-number-format/pull/80/files#diff-520385b0e4cf9fbd5592b5e80fb833f0R146

src/number_format.js
```diff
@@ -128,12 +129,12 @@ class NumberFormat extends React.Component {
...
     props = props || this.props;
     const {decimalSeparator, thousandSeparator} = this.getSeparators(props);
-    return num.replace(new RegExp(escapeRegExp(thousandSeparator || ''), 'g'), '').replace(decimalSeparator, '.');
+    return (num || '').replace(new RegExp(escapeRegExp(thousandSeparator || ''), 'g'), '').replace(decimalSeparator, '.');
   }
 
   getFloatValue(num, props) {
     props = props || this.props;
-    return parseFloat(this.getFloatString(num, props)) || 0;
+    return parseFloat(this.getFloatString(num || '', props)) || 0;
```
s-yadav: Here or operation is not required as it's already been done on getFloatString

src/number_format.js
```diff
@@ -142,7 +143,7 @@ class NumberFormat extends React.Component {
 
     let {value} = props;
 
-    if (format || value === undefined || value === '') return value;
+    if (format || !value) return value;
```
s-yadav: This will return in case value is 0 as well. It shouldn't

s-yadav:
```js
if (format || !(value || value === 0)) return value;
```

isnifer:
@s-yadav hope, that's all

src/number_format.js
```diff
@@ -84,6 +84,7 @@ const defaultProps = {
   suffix: '',
   allowNegative: true,
   type: 'text',
+  value: '',
```
s-yadav: Test failing because of this as if value is not provided we treat it as an uncontrolled component.

PR #87, opened 2017-09-17
prefix and suffix bug fix

geforcefan:
In formatInput, removePrefixAndSuffix should be called before anythings else, in order that the suffix or prefix may contain any number or signs (example suffix: "e-005 m/s²").


s-yadav:
It's fixed on v3

PR #91, opened 2017-09-28
Support for disabling the auto generated decimal values

klangenk:
Hi. I wanted to use the decimalPrecision property, so the user is not allowed to input more decimal places than defined. But I did not want to have the autogenerated x.00. I implemented a new prop 'fillDecimalZeroes' which is true on default.


s-yadav:
It's available now on v3

PR #111, opened 2017-10-28
Fix onFocus caret reset position bug

pandaiolo:
This bug appears on OSX Safari and Chrome, at least. Ref https://bugs.chromium.org/p/chromium/issues/detail?id=779328

v3 now corrects focus caret position, but relies on `event.target.selectionStart`, which is not immediatelly available due to this bug. The reported value is always `0` instead of mouse position. 

This PR offers a workaround to avoid caret position reset, until browser vendors push a fix,  and is based on `v3` branch:

- Add a test to fail on this bug
- Wrap `onFocus` event handler in a `setTimetout(0)`
- Fix other on focus test to work with async behaviour

Other fix: 
- Small naming typo fix (`caretPosition`)

### DIFF
--- a/src/number_format.js
+++ b/src/number_format.js
@@ -668,15 +668,20 @@ class NumberFormat extends React.Component {
   }
 
   onFocus(e: SyntheticInputEvent) {
-    const el = e.target;
-    const {selectionStart, value} = el;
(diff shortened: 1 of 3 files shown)


s-yadav:
Looks good. Thanks.

PR #119, opened 2017-11-14
Add disabled attribute

florianbepunkt:
Added disabled attribute to input. Should be a non breaking change.


s-yadav:
This is not required, as it supports all input/customInput props other than props defined on documentation. So disabled prop will work in the current version as well.

PR #123, opened 2017-11-22
Add optional autoFocus prop (fixes #122)

### LINKED ISSUE #122: Add autoFocus prop.
Just pass it through to input component!

crownz:


s-yadav:
This is not required, passing autoFocus without this will work. As any props other than defined in the doc are forwarded to input (Mentioned in the doc below options).

Issue #121, opened 2017-11-20
In number input if to press Ctrl+A and then type `.`, nothing happens 


khmelevskii:
My input with following props
```
            thousandSeparator: true,
            prefix: '$',
            decimalScale: 2,
```
Where there is a value in input and I press ctrl-a and then I want to type decimal number, I need to press `.` two times as it doesn't work from the first time.

Also you can check it on your demo page

6degreeshealth:
I have this problem as well. I expect that selected text is overwritten on keypress. But, when decimalScale is set, it isn't overwritten.

PR #126, opened 2017-11-27
Fix partial deletes

vincaslt:
I believe this component shouldn't change default behaviour dictated by html's <input> component. How it currently works: when you select all text in the component, and type a valid character, nothing happens (this is true for components that have spaces/characters as separators), or if you select part of the input, that contained separator, and press delete - nothing happens. It should always work the same way an input does - you press delete - part of the text selected is deleted, you select and type - text is overwritten.

This fixes: #121

### DIFF
--- a/src/number_format.js
+++ b/src/number_format.js
@@ -556,18 +556,6 @@ class NumberFormat extends React.Component {
       return value;
     }
 
-    const start = caretPos;
-    const lastValueParts = splitString(lastValue, caretPos);
-    const newValueParts = splitString(value, caretPos);
-    const deletedIndex = lastValueParts[1].lastIndexOf(newValueParts[1]);
-    const diff = deletedIndex !== -1 ? lastValueParts[1].substring(0, deletedIndex) : '';
-    const end = start + diff.length;
-
-    //if format got deleted reset the value to last value
-    if (this.checkIfFormatGotDeleted(start, end, lastValue)) {
-      value = lastValue;
-    }
-
     //for numbers check if beforeDecimal got deleted and there is nothing after decimal,
     //clear all numbers in such case while keeping the - sign
     if (!format) {
@@ -590,10 +578,10 @@ class NumberFormat extends React.Component {
     const {state, props} = this;
     const {isAllowed} = props;
     const lastValue = state.value || '';
-
+    
     /*Max of selectionStart and selectionEnd is taken for the patch of pixel and other mobile device caret bug*/
     const currentCaretPosition = Math.max(el.selectionStart, el.selectionEnd);
-
+    
     inputValue =  this.correctInputValue(currentCaretPosition, lastValue, inputValue);
 
     let formattedValue = this.formatInput(inputValue) || '';


s-yadav:
Did you run the test cases?
Partial delete is blocked for few reasons. As format pattern allows number within format/prefix/suffix. In such scenario, if the partial delete is applied it's hard to identify actual user input vs the format. For example, let's say the format is`+1-(##33)`, and user typed `+1-(4333)` and then remove `(4` resulting `+1-333`. Now in this string its hard to identify what is the part of format and what was user input.
So in the current implementation, it breaks the format in parts `+1-(`, `##`, `33)', and only removing hash part is allowed.  In this case, we know that what is the place for user input and deletion of that part can easily be covered.

However ctrl + a, and delete is allowed. #121 is some other bug which has to be fixed.

vincaslt:
ctrl+a delete is allowed, because it's explicitly checked. When you do ctrl+a 5 for example, nothing is entered, because that logic swallows the input, since it removes separators in the process. I get why you want to block partial delete, but that logic should be reconsidered.

vincaslt:
I was thinking how you can make partial delete work, and I think it may be possible if the value that we get we always try to conform to a format first, before removing formatting. E.g. in your suggested case it would be: format `+1-(##33)` and entered value is `+1-333)`. We can iterate over format, and conform the value to it like so:

# `+` is format and exists in value - we keep it. -> `+`
# `1` is format and exists in value, so we keep it -> `+1`
# `-` is format, and exists in value -> `+1-`
# `(` is format, but doesn't exist in value at the same position, so we correct the value, and check the same position in value in the next iteration (saving `(` in some list to ignore it later when we actually find it in value, if we find it) -> `+1-(`
# `#` is not format, so we take the number, previously we had invalid values, so now we check the same position in value again, it's not in ignored list (`(`), so -> `+1-(3`
# `#` again not format, so we take the next number `+1-(33`
# `3` is format, and our value is `3` too, so we assume it's formatting and take it -> `+1-(333`
# `3` is format, but our value is `)`, so we fix it and will check the same position in the next iteration -> `+1-(3333`
# `)` is format, and we find it's the same in value, so it's formatting there too, we take it and it's over -> `+1-(3333)`

I have tried implementing this algorithm, and it worked in general case. The problems I've encountered were: if you entered `11` in your case, value would be fixed to `+` at first, then the next number it found is `1` and it's the same as formatting, so it would be mistaken for formatting: `+1-(1_33)` instead of `+1-(1133)`. Another problem is with decimal numbers, they don't have formatting so naturally iterating over it is impossible, and the last one was negative numbers, but that might be unrelated, when negative number is fixed first, minus sign is taken for a number and it gets removed along, so one number gets clipped off. 

I can push my implementation to some branch on my fork and you can check it out, but if we solved those problems I mentioned it might actually be a feasible solution.

s-yadav:
> I can push my implementation to some branch on my fork and you can check it out, but if we solved those problems I mentioned it might actually be a feasible solution.

@vincaslt Sure that will help. 

The other case in which it might break is consider formatting +1-(##33)-## and the value to be +1-(4433)-33
Now if someone deletes `33)-`, user input 33 will be considered as format 33.

Issue #125, opened 2017-11-26
How to get ref to the input?


mbrevda:
I'm using this with material-ui like:

```js

class Amount extends Component {
  render() {
    return <NumberFormat {...this.props} decimalScale={2} />
  }
}

render() {
            <TextField
              className={classes.textField}
              fullWidth={true}
              label="Amount"
              name="amount"
              onChange={this.onAmountChange}
              required={true}
              InputProps={{
                startAdornment: (
                  <InputAdornment position="start">$</InputAdornment>
                ),
                inputRef: ref => this.trackRef('amount', ref),
                inputComponent: Amount
              }}
}
```

But I don't get a ref to the actual input. How can I get the ref?

Edit: it seem that the `omit` help function isn't copying a `ref` function if one is provided

s-yadav:
`ref` attributes are not passed as props to the component See (https://reactjs.org/warnings/special-props.html). They are internally used by react.

I guess in your Amount component you have to manually pass ref . Something like
```jsx
<NumberFormat ref={this.props. inputRef} {...resr} decimalScale={2} />
```
But in this case as well ref will not forward to actual input. I guess I also have to support something similar `inputRef` like TextField Component.

mbrevda:
>  I guess I also have to support something similar inputRef like TextField Component

Cool, thanks!

pvpacheco:
I've just submitted a pull request (#131) implementing the inputRef prop, hope it helps...

s-yadav:
This has been implemented on v3.1.0
https://github.com/s-yadav/react-number-format#getting-reference

mbrevda:
Whoa! This is awesome, thanks!

PR #131, opened 2017-12-05
allowing a ref to input element when displayType=input

pvpacheco:
I had an issue similar to #125 in a situation where I needed to implement a setfocus() call on the input element. 

So I've implemented the inputRef like TextField Component, allowing, for instance:

`<NumberFormat inputRef={(el) => {this.inputElement = el}} thousandSeparator={true} decimalScale={3} fixedDecimalScale={true} prefix="$" />`

Hope it helps. Cheers.


unaimian:
How to get the ref value? Nothing worked for me so far..

PR #139, opened 2017-12-12
Allow password type

pr0da:
Beside the prop type error message, I haven't noticed any issue with password type.
Possible use cases: cvc input.


s-yadav:
Why you need formatting on password type input? Anyway, the text is not visible there.  You should use normal input[type="password"] there.

PR #148, opened 2018-01-23
Handle Null Case

BukhariH:
We had issues when we upgraded to `3.1.3` with the library handling `null` as a value prop. 

Should be treated the same as undefined?


s-yadav:
This has been fixed on latest release. Closing it.

Issue #134, opened 2017-12-11
fixedDecimalScale=true & decimalScale=2 breaks overwriting selection


trbarrett:
Hi there,

Version 3.1.0

I just ran into this issue were users were unable to select the text and type in (or paste) a new one. Either the number just appears in-font of the existing number, or nothing happens. After trying a few different options I nailed it down to "fixedDecimalScale={true}"  combined with "decimalScale={2}" causing the issue. 

To reproduce:

1. Go to https://codepen.io/lendinvesttristian/pen/eyOEvm
2. Type in a number in the first input box
3. Select the number (double click or drag and select)
4. Type a number, or paste a number

If you try that in the second or third input in https://codepen.io/lendinvesttristian/pen/eyOEvm, then typing or pasting a number works fine.

Cheers,
Tristian Barrett

drinkwater99:
I have the same issue, is there any workaround ?

pavelKozlov:
Possible soltuion: https://github.com/s-yadav/react-number-format/pull/152.

s-yadav:
Fixed in the latest release. 3.1.4

starcraft0429:
Thanks for update.
Btw, still issue is .
when I insert 50.42( %)
and select all number.

if I insert all different number, it works well.
but I start with same number(for example I was gonna 53.42) it was same issue with previous.

Hope to fix this soon
Thanks again.

Issue #142, opened 2017-12-21
Cannot overwrite full number with fixedDecimalScale


gurnzbot:
When `fixedDecimalScale` is set to true and there is a value in the component, I am unable to overwrite the component's value entirely.

For example:
Initial component value: 4.50
Highlight/select the entire number
Type in a new number: 7

Result: Number does not change
Expected: 7.00

This throws off any users I've run across.  They expect the number to be overwritten.

businessnumbertwo:
Yes - I've come across this too, and can confirm what gurnzbot says above

pavelKozlov:
Possible solution: https://github.com/s-yadav/react-number-format/pull/152.

s-yadav:
Duplicate of #134

s-yadav:
Fixed in v3.1.4

johnny-carmona:
IK'm using version 4.0.6 and this is still happening.

PR #152, opened 2018-01-31
Insert value if entire range is selected

pavelKozlov:
When fixedDecimalScale is set to true and there is a value in the component, I am unable to overwrite the component's value entirely.

For example:
Initial component value: 4.50
Highlight/select the entire number
Type in a new number: 7

Result: Number does not change
Expected: 7.00

The description is taken from https://github.com/s-yadav/react-number-format/issues/142.
It also fixes https://github.com/s-yadav/react-number-format/issues/134.

### DIFF
--- a/src/number_format.js
+++ b/src/number_format.js
@@ -547,8 +547,7 @@ class NumberFormat extends React.Component {
     const diff = deletedIndex !== -1 ? lastValueParts[1].substring(0, deletedIndex) : '';
     const end = start + diff.length;
 
-    const isDeleteAll = lastValueParts[1].length === deletedIndex;
-
+    const isDeleteAll = newValueParts[1] === '';
     //if format got deleted reset the value to last value
     if (!isDeleteAll && this.checkIfFormatGotDeleted(start, end, lastValue)) {
       value = lastValue;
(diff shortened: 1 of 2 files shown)


src/number_format.js
```diff
@@ -547,8 +547,10 @@ class NumberFormat extends React.Component {
     const diff = deletedIndex !== -1 ? lastValueParts[1].substring(0, deletedIndex) : '';
     const end = start + diff.length;
 
+    const isDeleteAll = lastValueParts[1].length === deletedIndex;
```
s-yadav: Wouldn't it will mark isDeleteAll for the case where all text after caretPos is selected. but there is text before caret position

s-yadav:
Can you check as other text are failing

pavelKozlov:
Hi @s-yadav,

You're right that the check if entire input value is replaced is not 100% accurate.
So I changed it.
I have also tested with different formats and masks from example sections and it works ok for me.
Maybe you can find other cases.
I have also added a test to cover new case and run all existent tests to be sure my changes don't break current code.

s-yadav:
Fixed this on 3.1.4. Let me know if this still exist

PR #165, opened 2018-03-01
Fix caret moving to the end of input

a-x-:
on press LeftArrow while at the 0-th pos


s-yadav:
It's already been fixed on v3.2.0

a-x-:
Ok, thank you, I'll check it

PR #166, opened 2018-03-10
fix #145

### LINKED ISSUE #145: Pasting a number with decimal while decimal scale is set to 0 gives wrong formatted value
Setup
```jsx
<NumberFormat decimalScale={0} />
```
Paste 9.5 to it, formatted value should be 9. Currently it comes as 95 which is incorrect.

ad1992:
@s-yadav please look into this


src/number_format.js
```diff
@@ -164,6 +164,10 @@ class NumberFormat extends React.Component {
   getNumberRegex(g: boolean, ignoreDecimalSeparator?: boolean) {
     const {format, decimalScale} = this.props;
     const {decimalSeparator} = this.getSeparators();
+
+    if(decimalScale === 0) {
```
s-yadav: This will break if the number has a thousand separators

ad1992:
@s-yadav can you check now. I have made the fixes.

src/number_format.js
```diff
@@ -590,9 +590,18 @@ class NumberFormat extends React.Component {
...
-    const {isAllowed} = props;
+    const {isAllowed, decimalScale} = props;
     const lastValue = state.value || '';
 
+    /*
+    * get the valid numerical values only before the decimal
+    * separator when decimal scale is 0, issue #145
+    */
+
+    if(decimalScale === 0) {
```
s-yadav: You should also check for if format prop is not defined otherwise it will break things.

s-yadav:
looks good will merge.

PR #168, opened 2018-03-12
Allows additional decimal separators

Berardpi:
Adds props to specify additional decimal separators so that the user can type both '.' and ',' but it still get formatted with the decimalSeparator


anissaHireme:
any news on this issue ? is seems it was not merged

s-yadav:
General use case is to allow typing `.` as decimal separator while decimalSeparator is other than `.` . No need for allowedDecimalSeparator prop on v4. The above use case will work. 
Closing this PR.

PR #160, opened 2018-02-15
Allow empty string formatting by custom formatter

martinrogachevsky:

PR #169, opened 2018-03-13
 Add allowEmptyFormatting property that applies formatting to empty input

AM-NiceDay:
```
<NumberFormat format="+1 (###) ###-####" mask="_" allowEmptyFormatting/>
```

leads to input with value `+1 (___) ___-____`

This is a reproduction of https://github.com/s-yadav/react-number-format/pull/160 due to inactivity of previous contributor.


s-yadav:
Nice, this looks clean. I guess this takes care of #160 as well.

s-yadav:
There is an issue on this. if allowEmptyFormatting on focus caret position is wrong. I will fix that and merge

PR #194, opened 2018-06-23
Fix #181 - Move Babel transforms to dependencies

### LINKED ISSUE #181: build fails - unknown plugin - "transform-runtime" - 3.3.3  version
Unknown plugin "**transform-runtime**" specified in "... node_modules\\react-number-format\\.babelrc" at 2, attempted to resolve relative to "... node_modules\\react-number-format"

svicalifornia:
This change fixes issue #181 by moving the Babel transforms to the `dependencies`, where they will be available to .babelrc when compiling surrounding projects in non-development environments.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,9 +35,6 @@
     "babel-eslint": "^6.0.0",
     "babel-loader": "^6.2.4",
     "babel-plugin-react-transform": "^2.0.2",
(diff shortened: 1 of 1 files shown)


svicalifornia:
After making some changes to our webpack config, we've determined the problem may have been on our side, and this change may not be necessary.

PR #198, opened 2018-07-08
Fix input of negative number in case of isNumericString & decimalScale

andemerie:
I used a component similar to the following:
```
<NumberFormat
  value={this.state.value}
  isNumericString
  onValueChange={({value}) => this.setState({value})}
  decimalScale={2}
/>
```
And the combination of isNumericString and decimalScale props caused bug with entering negative number. When I entered minus sign, input immediately showed "-0" and this zero could not be removed using backspace.

### DIFF
--- a/test/library/input_numeric_format.spec.js
+++ b/test/library/input_numeric_format.spec.js
@@ -467,4 +467,28 @@ describe('Test NumberFormat as input with numeric format options', () => {
     expect(wrapper.state().value).toEqual('1');
   });
 
+  it(`should not add 0 after minus immediately after minus is entered in case isNumericString and
+    decimalScale props are passed`, () => {
+    class IssueExample extends React.Component {
+      constructor() {
+        super();
+        this.state = {
+          value: ''
+        };
+      }
+      render() {
+        return (
+          <NumberFormat
+            value={this.state.value}
+            isNumericString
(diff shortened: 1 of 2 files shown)


andemerie:
@s-yadav Hi there! How should we proceed on this PR?

s-yadav:
@andemerie Sorry, I missed this PR earlier. The changes looks good. Can you add a test case as well to cover this bug?

andemerie:
@s-yadav Done :)

test/library/input_numeric_format.spec.js
```diff
@@ -467,4 +467,28 @@ describe('Test NumberFormat as input with numeric format options', () => {
...
+            value={this.state.value}
+            isNumericString
+            onValueChange={({value}) => this.setState({value})}
+            decimalScale={2}
+          />
+        )
+      }
+    }
+    const wrapper = mount(<IssueExample/> );
+    simulateKeyInput(wrapper.find('input'), '-');
```
s-yadav: I don't think you need to create a class here.

```js
const wrapper = mount(<NumberFormat isNumericString decimalScale={2} />);
simulateKeyInput(wrapper.find('input'), '-', 0, 0);
```
This should be enough I guess.

s-yadav:
Made the required changes and merged

PR #200, opened 2018-07-16
added typings for NumberFormat component

horacehylee:
- added typings for NumberFormat component for better Typescript experience
- added types property in package.json for tsc to recognize declaration file
- updated README to use default import as the library is using module.exports

### DIFF
--- a/typings/number_format.d.ts
+++ b/typings/number_format.d.ts
@@ -1,0 +1,54 @@
+/// <reference types="react" />
+
+declare module "react-number-format" {
+  export interface NumberFormatState {
+    value?: string;
+    numAsString?: string;
+  }
+
+  export interface NumberFormatValues {
+    floatValue: number;
+    formattedValue: string;
+    value: string;
+  }
+
+  type FormatValueFunction = (formattedValue: string) => string;
+
+  export interface NumberFormatProps
+    extends React.HTMLAttributes<HTMLInputElement> {
+    thousandSeparator?: boolean | "," | "." | " ";
+    decimalSeparator?: boolean | "," | ".";
+    decimalScale?: number;
+    fixedDecimalScale?: boolean;
+    displayType?: "input" | "text";
+    prefix?: string;
+    suffix?: string;
+    format?: string | FormatValueFunction;
+    removeFormatting?: (formattedValue: string) => string;
+    mask?: string | string[];
+    value?: number | string;
+    isNumericString?: boolean;
+    customInput?: (inputComponent: React.ComponentType<any>) => void;
+    allowNegative?: boolean;
+    allowEmptyFormatting?: boolean;
+    onValueChange?: (
+      values: NumberFormatValues,
+      e: React.ChangeEvent<HTMLInputElement>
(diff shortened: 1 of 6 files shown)


s-yadav:
Great. I was waiting for someone to create this PR. 😃. Will check this by tonight.

horacehylee:
Having the type declaration file here, save the time to install an extra dev dependencies 😊. 
Great library BTW! 🎉🎉
Saved a lot of time from struggling with currency input.

typings/number_format.d.ts
```diff
@@ -0,0 +1,54 @@
...
+    floatValue: number;
+    formattedValue: string;
+    value: string;
+  }
+
+  type FormatValueFunction = (formattedValue: string) => string;
+
+  export interface NumberFormatProps
+    extends React.HTMLAttributes<HTMLInputElement> {
+    thousandSeparator?: boolean | "," | "." | " ";
```
s-yadav: thousandSeparator can be boolean or any string

typings/number_format.d.ts
```diff
@@ -0,0 +1,54 @@
...
+    decimalSeparator?: boolean | "," | ".";
```
s-yadav: Same

typings/number_format.d.ts
```diff
@@ -0,0 +1,54 @@
...
+    numAsString?: string;
+  }
+
+  export interface NumberFormatValues {
+    floatValue: number;
+    formattedValue: string;
+    value: string;
+  }
+
+  type FormatValueFunction = (formattedValue: string) => string;
```
s-yadav: It should be 
```js
type FormatValueFunction = (inputValue: string) => string;
```

typings/number_format.d.ts
```diff
@@ -0,0 +1,54 @@
...
+    removeFormatting?: (formattedValue: string) => string;
+    mask?: string | string[];
+    value?: number | string;
+    isNumericString?: boolean;
+    customInput?: (inputComponent: React.ComponentType<any>) => void;
+    allowNegative?: boolean;
+    allowEmptyFormatting?: boolean;
+    onValueChange?: (
+      values: NumberFormatValues,
+      e: React.ChangeEvent<HTMLInputElement>
```
s-yadav: e will not be always a ChangeEvent, It can be blur event as well. Some transformation happens on blur as well which calls onValueChange

horacehylee: Should I set e to any or something similar to SyntheticInputEvent?

s-yadav: SyntheticInputEvent will be better. If its not available you can change it to something like `e: React.ChangeEvent<HTMLInputElement> | React.BlurEvent<HTMLInputElement>`

horacehylee:
@s-yadav I have made those changes.
I defined a SyntheticInputEvent by referencing https://github.com/facebook/flow/blob/v0.49.1/lib/react.js#L310

s-yadav:
Looks good. I will merge

PR #202, opened 2018-07-18
update custom input type def

danielwong2268:


typings/number_format.d.ts
```diff
@@ -34,7 +34,7 @@ declare module "react-number-format" {
     mask?: string | string[];
     value?: number | string;
     isNumericString?: boolean;
-    customInput?: (inputComponent: React.ComponentType<any>) => void;
+    customInput?: React.ReactElement,
```
s-yadav: customInput is a React component class or a function reference (ex : TextField). `React.ReactElement` will be valid for instance (ex: <TextField />). 
Though `(inputComponent: React.ComponentType<any>) => void;` is also not correct

s-yadav:
Looks like this has been already fixed

Issue #196, opened 2018-07-03
Allow characters/custom functions to set the values


Princezhm:
I was wondering if there is a way to set a pre evaluation function to evaluate some inputs, for example I was thinking on an user which enters "1K" and if it is a valid entry a function will be able to return the actual value (1000).

if null is returned, it is not valid or it doesnt have a valid mapping.

```javascript
function preEvaluator(value){
    if(value === "1k") return 1000;

    return null
}
......
......

<NumberFormat preEvaluate={preEvaluator} ..... />
```

is there a way to accomplish this now? because I don't see a way to do it.

awesor:
I was looking for the exact same thing.  Unfortunately, I don't think that any alpha characters are allowed, unless they're passed via the suffix.  Even then, the user cannot choose their own suffix ("K", "M" - millions, "B"- billions).

Definitely would love to see alpha-chars supported!

s-yadav:
@Princezhm @awesor preEvaluate can be a option, where we can allow preEvaluate to transform number before NumberFormat do any transformation on it. The only problem with preEvaluate is how we will calculate correct caret position after it.

I can think of this as an experimental props with limited use case. 

Well, I am open for discussion on this feature on what all cases can come with this prop and what we can support.

Princezhm:
hi @s-yadav I want to contribute to your repo, so I started to see what we can do, for example we got this options, maybe @awesor can add some cases:

- have a super smart input to modify the different inputs, for example typing `"x"` on top of `1.233|452`, which will result on `1.233x|452`, but what can we do with this? I don't get a case for this.
- just have let the user to validate the key that was pressed and return null if nothing needs to be done, in any other case return a number/float to multiply the current value to the new one. for example if the user has `1.22` and in our `customFunction` we receive `k` we can return `1024` (obviously this is up to the user) and therefore we will multiply `1.22 * 1024 =1249.28 ` and we apply the correct formatting to the `1249.28` number, which is the final result.  one thing to notice with this is that if we do it this way, only allowing the final key to be the one to use as a precalculator, we don't need to worry about the caret position (correct me if I'm wrong), because it will always end at the end of the value, in the example it goes from `1.22|` to `1249.28|`

so, the custom transformation function could be something like:

```javascript
/**
   *
   * @param { string } key: key that was pressed by the user, this function will be called only on special keys.
   * @returns { Number / Null } Number to multiply the value or null if the user doesnt allow the character
   */
  customTransformation(key) {
[...]

PR #205, opened 2018-07-20
Support a preEvaluate function to modify values

Princezhm:
- add support to a pre evalute function
- check and correct the caret when used
- if no response, continue the normal flow
- added a example for the new functionality

as an example of the function that should be used to format the numbers we have this.

```javascript
evaluate(value, unformatted) { 
    if (value.endsWith("k")) { 
      let v = parseFloat(value) * 1000; 
      return v.toString(); 
    } else if (value.endsWith("m")) { 
      return (parseFloat(value) * 1000 * 1000).toString(); 
    } else if(value.endsWith('n')){ 
      return (parseFloat(value) * 0.01).toString(); 
    } 
    return null; 
  } 
```

and the usage should be:

```javascript
render(){
    return (<NumberFormat preEvaluate={this.evaluate}/>);
}
```


Princezhm:
@s-yadav let me know if you need to do any changes.

@awesor related to https://github.com/s-yadav/react-number-format/issues/196

pnewhook:
Can this be merged? I have the exact same requirement and until this library can support special alpha characters I'm stuck with my own custom solution.

linhnq-1077 (approved):

Issue #220, opened 2018-09-11
getInputRef type mismatch with using TextField material-ui.


IgnorantCoder:
Hello.
I am using this module with material-ui.

I got the compile error in using it in TypeScript.
I found out type mismatch between getInputRef and inputRef that passed from material-ui.
I want to make type restriction of getInputRef lax for solving this problem.

Problem:
```
        <NumberFormat
            getInputRef={inputRef} 
            value={value as number}
            onValueChange={(values) => {
                if (onChange === undefined) {
                    return;
                }

                const event = {
                    target: {
                        value: values.value,
                    },
                } as React.ChangeEvent<HTMLInputElement>;
                onChange(event);
            }}
        />
```

error:
```
TS2326: Types of property 'getInputRef' are incompatible.
  Type 'string | ((instance: any) => any) | RefObject<any> | undefined' is not assignable to type '((el: HTMLInputElement) => void) | undefined'.
  Type 'string' is not assignable to type '((el: HTMLInputElement) => void) | undefined'.
```

Workarround:
```
        <NumberFormat
            getInputRef={inputRef as (el: HTMLInputElement) => void}
            value={value as number}
            onValueChange={(values) => {
                if (onChange === undefined) {
                    return;
                }

                const event = {
                    target: {
                        value: values.value,
                    },
                } as React.ChangeEvent<HTMLInputElement>;
                onChange(event);
            }}
        />
```

PR #221, opened 2018-09-11
Update number_format.d.ts

IgnorantCoder:
Relates to #220

### DIFF
--- a/typings/number_format.d.ts
+++ b/typings/number_format.d.ts
@@ -52,7 +52,7 @@ declare module "react-number-format" {
     type?: "text" | "tel";
     isAllowed?: (values: NumberFormatValues) => boolean;
     renderText?: (formattedValue: string) => React.ReactNode;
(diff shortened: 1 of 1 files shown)


typings/number_format.d.ts
```diff
@@ -52,7 +52,7 @@ declare module "react-number-format" {
     type?: "text" | "tel";
     isAllowed?: (values: NumberFormatValues) => boolean;
     renderText?: (formattedValue: string) => React.ReactNode;
-    getInputRef?: (el: HTMLInputElement) => void;
+    getInputRef?: ((el: HTMLInputElement) => void) | React.Ref<any> | string;
```
s-yadav: Passing getInputRef as string will break. Why do we want to keep that as option ?

IgnorantCoder: Oh sorry.
I think supporting string is not good, too.
I want to support React.Ref<any> only.

s-yadav:
Looks good. Can you squash your commits ?

s-yadav:
Merged now. Thanks.

Issue #253, opened 2018-11-26
Malicious library flatmap-stream is a dependency because of nodemon


jpamarohorta:
Malicious library `flatmap-stream` is a dependency. This library is malicious and it's not completely clear what it does as it is explained [here](https://github.com/dominictarr/event-stream/issues/116).

The problem lies with `nodemon` that is dependecy of this project. It has then the following nested dependencies: `pstree.remy` -> `ps-tree` -> `event-stream` -> `flatmap-stream`.

Given that nobody knows yet what this code really does, this can be problematic.

s-yadav:
We can remove nodemon safely from dependency as it's no longer required.

jpamarohorta:
Ok thanks. Created a PR that removes it.

s-yadav:
Pushed the fix on 4.0.5

PR #254, opened 2018-11-27
nodemon removed as a dependency

jpamarohorta:
nodemon removed as a dependency as it is not needed anymore and it was affected by a nested malicious package.

[Related issue](https://github.com/s-yadav/react-number-format/issues/253)


s-yadav:
Didn't saw this PR. Have made the changes on latest release.

PR #275, opened 2019-01-23
Allow removal of format if user has a selection

dsturley:
If a user selects multiple characters which spans a prefix and then attempts to delete or change their selection, the input reverts to previous value.

https://codepen.io/s-yadav/pen/bpKNMa

## Deleting
Steps to reproduce:
1. Move to "Prefix and thousand separator : Format currency in input" input
2. Using mouse; select '$1'
3. Hit delete/backspace

Expected:
Input updates to '$23,967'

Actual:
Input reverts to '$123,967'

## Editing

Steps to reproduce:
1. Move to "Prefix and thousand separator : Format currency in input" input
2. Using mouse; select '$1'
3. Hit '2'

Expected:
Input updates to '$223,967'

Actual:
Input reverts to '$123,967'

### DIFF
--- a/src/number_format.js
+++ b/src/number_format.js
@@ -628,8 +628,9 @@ class NumberFormat extends React.Component {
       return value;
     }
 
-    //if format got deleted reset the value to last value
-    if (this.checkIfFormatGotDeleted(start, end, lastValue)) {
+    //if format got deleted reset the value to last value, unless the user has
+    //a selection spanning multiple characters
+    if (this.checkIfFormatGotDeleted(start, end, lastValue) && start === end) {
       value = lastValue;
     }
 


src/number_format.js
```diff
@@ -628,8 +628,9 @@ class NumberFormat extends React.Component {
       return value;
     }
 
-    //if format got deleted reset the value to last value
-    if (this.checkIfFormatGotDeleted(start, end, lastValue)) {
+    //if format got deleted reset the value to last value, unless the user has
+    //a selection spanning multiple characters
+    if (this.checkIfFormatGotDeleted(start, end, lastValue) && start === end) {
```
s-yadav: start === end will always be false.

s-yadav:
This was a intentional check to resolve lot of other bugs, as the prefix and suffix or format can have number into it and if partial delete is allowed it will be hard to figure out what was part of formatting characters vs whats the part of actual input.

s-yadav:
I will see if I can add few more specific checks. But I will keep this as won't fix PR, bug for now.

Issue #277, opened 2019-01-30
Changing value and suffix at the same time causes Maximum update depth exceeded error


AlminaHaskic:
I'm using react-number-format in combination with Formik and material-ui. I have a use case when I need to call Formiks resetForm and it updates value and a suffix at the same time. E.g. (100 m to 0 ft ). 

`const NumberFormatCustom = (props: *): NumberFormat => {
  const { inputRef, onChange, unit, ...other } = props;
  return (
    <NumberFormat
      {...other}
      getInputRef={inputRef}
      onValueChange={values => {
        onChange({
          target: {
            value: values.value,
            name: other.name,
          },
        });
      }}
      suffix={unit}
    />
  );
};`

NumberFormat is passed though InputProps to TextField.

`            InputProps={{ inputComponent: NumberFormatCustom }}
`

This is the console log that shows Maximum update depth exceeded error.

![screen shot 2019-01-30 at 5 15 00 pm](https://user-images.githubusercontent.com/22662973/51995251-a4d96500-24b2-11e9-8929-23e2689c2469.png)

rykon:
I was getting something similar -- "Maximum update depth..." message when a phone number field was autofilled by Chrome. 

I updated the `onValueChange()` function to check for the existence of the `values` object as well as the `values.formattedValue` key before calling `onChange`. This appears to have solved my problem. Hope this helps.

AlminaHaskic:
@rykon I did the same few days ago and it helps. Thank you. But still, it should be fixed in react-number-format.

rykon:
It looks like checking to see whether or not `formattedValue` is truthy before calling `onValueChange()` on [line #695](https://github.com/s-yadav/react-number-format/blob/master/src/number_format.js#L695) would fix the problem.

s-yadav:
checking formattedValue is truthy before calling onValueChange is not correct as when user press removes all the value in the input. The empty value has to be passed to the onValueChange method. 

Will check for the root cause.

s-yadav:
@AlminaHaskic @rykon Changing value and suffix at the same time works on the normal case. Can you create a codesandbox (https://codesandbox.io) with failing case. So it will be easier to debug.

morloy:
I think I found a workaround by using
```
value={this.state.value || undefined}
```

This briefly switches the component to uncontrolled if there’s now valid input and breaks the infinite loop.

wesleyguirra:
For me the problema is happening when I set name and value at onChange key:

[...]

PR #280, opened 2019-02-12
Issue #277: Check `formattedValue` before calling `onValueChange()`

rykon:
This addresses the "maximum update depth exceeded" issue reported in Issue #277 .

### DIFF
--- a/src/number_format.js
+++ b/src/number_format.js
@@ -692,7 +692,10 @@ class NumberFormat extends React.Component {
     //update state if value is changed
     if (formattedValue !== lastValue) {
       this.setState({value : formattedValue, numAsString}, () => {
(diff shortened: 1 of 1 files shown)


src/number_format.js
```diff
@@ -692,7 +692,10 @@ class NumberFormat extends React.Component {
     //update state if value is changed
     if (formattedValue !== lastValue) {
       this.setState({value : formattedValue, numAsString}, () => {
-        onValueChange(this.getValueObject(formattedValue, numAsString));
+        //prevent a "maximum update depth exceeded" error in React (issue #277)
+        if (formattedValue) {
```
s-yadav: This will break when user press backspace and remove everything on the input. The empty value has to be passed to the onValueChange.

PR #281, opened 2019-02-13
Use new prop-types elementType for customInput

### LINKED ISSUE #272: customInput failed prop type with styled-component passed
Hi, sorry but I got this error when I try to pass a styled-component

```
import styled from 'styled-components';

const StyledInput = styled((props) => (
  <input {...props} />
))`
  // my stylings...
`

```
to the customInput prop. The error that I got is the following

> Failed prop type: Invalid prop `customInput` of type `object` supplied to `NumberFormat`, expected `function`.

However the app works just fine.
What's the cause of this if I may get a better understanding? Thank you in advance!

qisoft:
Update prop-types to ^15.7.0
Use new elementType to validate customInput (https://github.com/facebook/prop-types/issues/200)
This should fix #272


s-yadav:
@qisoft The new prop-type version outputs a lot of extra bits on distribution files. Can you check what causes that.?

alextiley:
Any update on this one? We're applying a hack in our codebase to workaround this issue and it would be great to get it fixed.

s-yadav:
I have merged #313 as it points to the latest propt-types. I am checking why the new prototype increasing dist size. Will make a release after that.

PR #285, opened 2019-02-19
Added a disabled prop to the input props to allow <input> to be disabled

mmnavarr:
NumberFormat component does not currently allow for the <input> element to be disabled. This PR contains an additional 'disabled' prop that disables the input field.


s-yadav:
Doesn't just passing disable on NumberFormat works. NumberFormat passes all extra props to the actual element. So all props which an Input supports are supported by react-number-format by default.

nathanshelly:
I can confirm that I experience the same issue in a large codebase I'm working in though passed disabled works in a simple `create-react-app` example. Unfortunately not sure exactly what in the larger codebase is breaking this behavior.

I know that's not particularly actionable just wanted to clarify that it seems to be more complicated than completely broken or completely fine.

s-yadav:
@nathanshelly Can you share some example where passing disable is not working.

nathanshelly:
I unfortunately can't, it's code from my company that I can't share. And don't understand well enough what's causing the issue to create a minimum repo (sorry for not being particularly helpful :( )

PR #286, opened 2019-02-27
Update number_format.js

dobre-robert-marius:


s-yadav:
Instead of this API change, I would suggest to use ref to access the instance method.

PR #291, opened 2019-03-09
resolves #252

### LINKED ISSUE #252: Add support for using parentheses instead of '-' for negative numbers
It's very common in financial applications to express a negative number as ($111,222) as opposed to -$111,222.  Could we add a prop for this?

SamKirkland:
Add support for using parentheses instead of '-' for negative numbers

Sudhanshu this library is awesome :)


src/number_format.js
```diff
@@ -78,6 +78,7 @@ const defaultProps = {
   onBlur: noop,
   isAllowed: returnTrue,
   getInputRef: noop,
+  accounting: false,
```
s-yadav: Define it on proptype as well. Also I think the prop name should be `negationFormat` whose value can be 'parentheses | minus'  default to minus.

src/number_format.js
```diff
@@ -164,14 +165,14 @@ class NumberFormat extends React.Component {
 
     //remove negation for regex check
     const hasNegation = num[0] === '-';
-    if(hasNegation) num = num.replace('-', '');
+    if (hasNegation) num = num.replace('-', '');
```
s-yadav: We will have to write the logicfor removing the parentheses as well.

src/number_format.js
```diff
@@ -181,7 +182,7 @@ class NumberFormat extends React.Component {
     }
 
     //add negation back
-    if(hasNegation) num = '-' + num;
+    if (hasNegation) num = '-' + num;
```
s-yadav: Float string should have always minus sign instead of brackets. Bracket should be only formatting detail.

src/number_format.js
```diff
@@ -277,8 +278,16 @@ class NumberFormat extends React.Component {
 
     //in case of format as number limit between prefix and suffix
     if (!format) {
-      const hasNegation = value[0] === '-';
-      return clamp(caretPos, prefix.length + (hasNegation ? 1 : 0), value.length - suffix.length);
+      let hasNegation;
+      let negationLength = 0;
+      if (accounting) {
+        hasNegation = value[0] === '(';
```
s-yadav: We should also check if the number ends with ).

src/number_format.js
```diff
@@ -362,20 +371,30 @@ class NumberFormat extends React.Component {
...
+    const {format, prefix, suffix, accounting} = this.props;
 
     //remove prefix and suffix
     if (!format && val) {
-      const isNegative = val[0] === '-';
+      let isNegative = val[0] === '-';
+      let stringEnd = val.length;
+
+      if (accounting) {
+        isNegative = val[0] === '(';
```
s-yadav: For is negative check we should check the last character as well

src/number_format.js
```diff
@@ -362,20 +371,30 @@ class NumberFormat extends React.Component {
...
-      if (isNegative) val = val.substring(1, val.length);
+      if (isNegative) val = val.substring(1, stringEnd);
 
       //remove prefix
       val = prefix && val.indexOf(prefix) === 0 ? val.substring(prefix.length, val.length) : val;
 
       //remove suffix
-      const suffixLastIndex = val.lastIndexOf(suffix);
+      let suffixLastIndex = val.lastIndexOf(suffix);
+      if (accounting && isNegative) {
```
s-yadav: You need to handle adding parentheses back. 
We do it on 
```js
if (isNegative) val = '-' + val;
```

src/number_format.js
```diff
@@ -804,8 +835,9 @@ class NumberFormat extends React.Component {
         e.persist();
 
         this.updateValue({formattedValue: newValue, caretPos: newCaretPosition, input: el});
-      } else if (!negativeRegex.test(value[expectedCaretPosition])) {
-        while (!numRegex.test(value[newCaretPosition - 1]) && newCaretPosition > leftBound){ newCaretPosition--; }
+      }
+      else if (!negativeRegex.test(value[expectedCaretPosition])) {
```
s-yadav: Parentheses negation should be handled here as well.

src/number_format.js
SamKirkland: removePrefixAndSuffix will remove the parentheses so getFloatString should never have parentheses, instead the string will be a valid floating point value.

src/number_format.js
SamKirkland: I agree.
I've confirmed this is the case. I have the parentheses removal logic in removePrefixAndSuffix so the brackets should never get to the underlying functions, they'll be removed/added in the removeFormatting and formatNumString functions

src/number_format.js
SamKirkland: Good point

src/number_format.js
SamKirkland: I avoided that approach so the number remains a valid floating point number when passed to inner functions.
The adding/removing logic for parentheses is in removePrefixAndSuffix and formatNumString.
I think this is the right approach so the underlying logic remains the same between the two, that way this setting just adds an automatic prefix/suffix to the number while removing the - when formatting the output.

README.md
```diff
@@ -213,14 +213,14 @@ You can easily extend your custom input with number format. But custom input sho
...
-  <NumberFormat customInput={TextField} format="#### #### #### ####"/>
+  <NumberFormat customInput={TextField} id="customInput" format="#### #### #### ####"/>
 ```
 ![Screencast example](https://media.giphy.com/media/3og0IH0LJhIQWFxztC/giphy.gif)
 
 **Passing custom input props**
 All custom input props and number input props are passed together.
 ```jsx
-  <NumberFormat hintText="Some placeholder" value={this.state.card} customInput={TextField} format="#### #### #### ####"/>
+  <NumberFormat hintText="Some placeholder" value={this.state.card} customInput={TextField} id="customInput" format="#### #### #### ####"/>
```
SamKirkland: Note: I was getting errors material-ui, id is required.

example/src/index.js
```diff
@@ -177,7 +181,7 @@ class App extends React.Component {
           <h3>
             Custom input : Format credit card number
           </h3>
-          <NumberFormat customInput={TextField} format="#### #### #### ####" />
+          <NumberFormat customInput={TextField} id="customInput" format="#### #### #### ####" />
```
SamKirkland: Same issue with material-ui

SamKirkland:
Is there anything else I need to do from my end to get this PR merged?

Thanks.

package-lock.json
```diff
@@ -0,0 +1,10411 @@
+{
```
s-yadav: Can you remove this file.? This module use yarn and it already have a yarn.lock file. Instead of npm install you can do yarn install.

src/utils.js
```diff
@@ -12,6 +12,16 @@ export function escapeRegExp(str: string) {
   return str.replace(/[-[\]/{}()*+?.\\^$|]/g, "\\$&");
 }
 
+export function negativeRegExp(negationFormat: string) {
+  switch (negationFormat) {
+    case 'parentheses':
+      return /\(/;
+    case 'negative':
```
s-yadav: The prop options are `minus` and `parentheses `. 
Also as we just have two options for negationFormat and minus is default, also we have added type check for that. Using ternary can reduce the overall size.
```js
export function negativeRegExp(negationFormat: string) {
  return negationFormat === 'parentheses' ? /\(/ : /-/
}
```

src/utils.js
```diff
@@ -24,6 +34,26 @@ export function getThousandsGroupRegex(thousandsGroupStyle: string) {
   }
 }
 
+export function getNegationPrefixSymbol(negationFormat: string) {
+  switch (negationFormat) {
```
s-yadav: Same we can use ternary here.

src/utils.js
```diff
@@ -24,6 +34,26 @@ export function getThousandsGroupRegex(thousandsGroupStyle: string) {
...
+    case 'parentheses':
+      return '(';
+    case 'negative':
+    default:
+      return '-';
+  }
+}
+
+export function getNegationSuffixSymbol(negationFormat: string) {
+  switch (negationFormat) {
```
s-yadav: ^^

test/library/input.spec.js
```diff
@@ -313,5 +313,13 @@ describe('NumberFormat as input', () => {
         shallow(<NumberFormat format="#### #### ####" mask={['D', 'D', 'M', '1', '2', 'Y', 'Y', 'Y']}/>)
       }).toThrow()
     })
-  })
+  });
+
+  it('should allow negationFormat to be parentheses', () => {
```
s-yadav: The spec statement is not very clear for the thing you are testing. You are only checking for empty/NAN values.

test/library/input_numeric_format.spec.js
```diff
@@ -48,6 +48,17 @@ describe('Test NumberFormat as input with numeric format options', () => {
     expect(wrapper.state().value).toEqual('-$123.55');
   });
 
+  it('supports negative numbers when using parentheses negationFormat', () => {
+    const wrapper = shallow(<NumberFormat thousandSeparator={true} prefix={'$'} negationFormat={'parentheses'} />);
+
+    simulateKeyInput(wrapper.find('input'), '-', 0);
+
+    expect(wrapper.state().value).toEqual('-');
```
s-yadav: Is this expected that on typing `-` first, it will be `-` and after typing number, it will change to parenthesis. I will see what is the expected behavior on a google sheet or MS Excel.

s-yadav: I feel whenever we press -, it should convert it to `()` even though there are no numbers yet

s-yadav:
> Is there anything else I need to do from my end to get this PR merged?
> 
> Thanks.

Sorry for the late review. Was quite busy with other things lately. I have added a few comments. 
Along with that.

1. Remove package-lock.json. We already have yarn.lock file.
2. Fix the conflicts.

package-lock.json
SamKirkland: Done!

src/utils.js
SamKirkland: Good catch, fixed.

test/library/input_numeric_format.spec.js
SamKirkland: Good point, I like it that way better too.
Done!

SamKirkland:
@s-yadav Let me know if you see anything else. Thanks.

s-yadav:
One bug I found, if you type - on the empty field (with parentheses negationFormat) the cursor goes on last. After that, I can not come in between the brackets. And then if I type any number, parenthesis go away.  
https://www.dropbox.com/s/x8hbdg21nke57b1/Screen%20Recording%202019-06-13%20at%2011.17.14%20AM.mov?dl=0

src/number_format.js
```diff
@@ -277,8 +282,18 @@ class NumberFormat extends React.Component {
 
     //in case of format as number limit between prefix and suffix
     if (!format) {
-      const hasNegation = value[0] === '-';
-      return clamp(caretPos, prefix.length + (hasNegation ? 1 : 0), value.length - suffix.length);
+      let hasNegation;
+      let negationLength = 0;
+      if (negationFormat === 'parentheses') {
+        hasNegation = value[0] === '(' && value[value.length - 1] === ')';
```
s-yadav: This logic is duplicate. Can we move it to util, and use that. 
```js
function getNegationMeta(value, negationFormat){
  let hasNegation;
  if (negationFormat === 'parentheses') {
    hasNegation = value[0] === '(' && value[value.length - 1] === ')';
  } else {
    hasNegation = value[0] === '-';
  }
  return {
     hasNagation,
     negationSuffixLegth: negationFormat === 'parentheses' && hasNegation ? 1 : 0,
  };
}
```
And this logic can be simple
```js
const {hasNagation, negationSuffixLegth} = getNegationMeta(value, negationFormat);
```

src/number_format.js
```diff
@@ -277,8 +282,18 @@ class NumberFormat extends React.Component {
 
     //in case of format as number limit between prefix and suffix
     if (!format) {
-      const hasNegation = value[0] === '-';
-      return clamp(caretPos, prefix.length + (hasNegation ? 1 : 0), value.length - suffix.length);
+      let hasNegation;
+      let negationLength = 0;
```
s-yadav: Change this variable name to negationSuffixLegth

src/number_format.js
```diff
@@ -360,22 +375,34 @@ class NumberFormat extends React.Component {
...
-  /** methods to remove formattting **/
+  /** methods to remove formatting **/
   removePrefixAndSuffix(val: string) {
-    const {format, prefix, suffix} = this.props;
+    const {format, prefix, suffix, negationFormat} = this.props;
 
     //remove prefix and suffix
     if (!format && val) {
-      const isNegative = val[0] === '-';
+      let isNegative = val[0] === '-';
```
s-yadav: You can use the same method here.
```js
const {hasNagation, negationSuffixLegth} = getNegationMeta(value, negationFormat);
//remove negation sign
if (isNegative) val = val.substring(1, val.length - negationSuffixLegth );
```

src/number_format.js
```diff
@@ -360,22 +375,34 @@ class NumberFormat extends React.Component {
...
+      if (isNegative) val = val.substring(1, stringEnd);
 
       //remove prefix
       val = prefix && val.indexOf(prefix) === 0 ? val.substring(prefix.length, val.length) : val;
 
       //remove suffix
-      const suffixLastIndex = val.lastIndexOf(suffix);
+      let suffixLastIndex = val.lastIndexOf(suffix);
+      if (negationFormat === 'parentheses' && isNegative) {
+        suffixLastIndex = suffixLastIndex - 1;
```
s-yadav: This logic looks incorrect, as you have already removed the parenthesis. I don't think we need to do suffixLastIndex - 1 here. Moreover, even though you wouldn't have removed parenthesis, lastIndexOf(suffix) will always give the correct value.

src/number_format.js
```diff
@@ -548,7 +579,13 @@ class NumberFormat extends React.Component {
     value = value.replace(/-/g, '');
 
     if (hasNegation && !removeNegation && allowNegative) {
-      value = '-' + value;
+      switch (negationFormat) {
```
s-yadav: Instead of switch, this would be shorter. As there are only two value of negationFormat;
value = negationFormat === 'parentheses' ? `(${value})` : `-${value}`;
:

src/number_format.js
```diff
@@ -797,14 +834,21 @@ class NumberFormat extends React.Component {
       negative value while the cursor position is after prefix. We can't handle it on onChange because
       we will not have any information of keyPress
       */
-      if (selectionStart <= leftBound + 1 && value[0] === '-' && typeof format === 'undefined') {
-        const newValue = value.substring(1);
+      const negativeValue = getNegationPrefixSymbol(negationFormat);
+
+      if (selectionStart <= leftBound + 1 && value[0] === negativeValue && typeof format === 'undefined') {
```
s-yadav: We can use the utility method here as well. 
```js
const {hasNagation, negationSuffixLegth} = getNegationMeta(value, negationFormat);
if (selectionStart <= leftBound + 1 && hasNagation && typeof format === 'undefined') {
  const newValue = value.substring(1, value.length - negationSuffixLegth);
```

test/library/format_as_text.spec.js
```diff
@@ -10,6 +10,16 @@ describe('NumberFormat as text', () => {
     expect(wrapper.find('span').text()).toEqual('$2,456,981');
   });
 
+  it('should format numbers to negative (type minus) currency', () => {
+    const wrapper = shallow(<NumberFormat value={-2456981} displayType={'text'} thousandSeparator={true} prefix={'$'} />);
```
s-yadav:
```suggestion
    const wrapper = shallow(<NumberFormat value={-2456981} displayType="text" thousandSeparator={true} prefix="$" />);
```

test/library/format_as_text.spec.js
```diff
@@ -10,6 +10,16 @@ describe('NumberFormat as text', () => {
...
+    expect(wrapper.find('span').text()).toEqual('-$2,456,981');
+  });
+
+  it('should format numbers to negative (type parentheses) currency', () => {
+    const wrapper = shallow(<NumberFormat value={-2456981} displayType={'text'} thousandSeparator={true} prefix={'$'} negationFormat={'parentheses'} />);
```
s-yadav:
```suggestion
    const wrapper = shallow(<NumberFormat value={-2456981} displayType="text" thousandSeparator={true} prefix="$" negationFormat="parentheses" />);
```

test/library/input.spec.js
```diff
@@ -113,7 +113,7 @@ describe('NumberFormat as input', () => {
       )
     }
 
-    const wrapper = mount(<WrapperComponent customInput={TextField} thousandSeparator={'.'} decimalSeparator={','}/>);
+    const wrapper = mount(<WrapperComponent customInput={TextField} id="customInput" thousandSeparator={'.'} decimalSeparator={','}/>);
```
s-yadav:
```suggestion
    const wrapper = mount(<WrapperComponent customInput={TextField} id="customInput" thousandSeparator="." decimalSeparator=","/>);
```

test/library/input_numeric_format.spec.js
```diff
@@ -48,6 +48,17 @@ describe('Test NumberFormat as input with numeric format options', () => {
...
+    const wrapper = shallow(<NumberFormat thousandSeparator={true} prefix={'$'} negationFormat={'parentheses'} />);
+
+    simulateKeyInput(wrapper.find('input'), '-', 0);
+
+    expect(wrapper.state().value).toEqual('()');
+
+    simulateKeyInput(wrapper.find('input'), '123.55', 1);
+
+    expect(wrapper.state().value).toEqual('($123.55)');
+  });
```
s-yadav: Can you also add spec for the failing case I have sent screen recording for?

test/library/input_numeric_format.spec.js
```diff
@@ -48,6 +48,17 @@ describe('Test NumberFormat as input with numeric format options', () => {
     expect(wrapper.state().value).toEqual('-$123.55');
   });
 
+  it('supports negative numbers when using parentheses negationFormat', () => {
+    const wrapper = shallow(<NumberFormat thousandSeparator={true} prefix={'$'} negationFormat={'parentheses'} />);
```
s-yadav:
```suggestion
    const wrapper = shallow(<NumberFormat thousandSeparator={true} prefix="$" negationFormat="parentheses" />);
```

test/library/input_numeric_format.spec.js
```diff
@@ -64,8 +75,23 @@ describe('Test NumberFormat as input with numeric format options', () => {
     expect(wrapper.state().value).toEqual('');
   });
 
+  it('removes negation when double negation is done when using parentheses negationFormat', () => {
+    const wrapper = shallow(<NumberFormat thousandSeparator={true} prefix={'$'} value={-2456981.89} negationFormat={'parentheses'} />);
```
s-yadav:
```suggestion
    const wrapper = shallow(<NumberFormat thousandSeparator={true} prefix="$" value={-2456981.89} negationFormat="parentheses" />);
```

alex-woodall-premierfoods:
This seems to have gone a bit quiet, be a shame to give up on it, I could do with this rather than hacking it manually using renderText

briandelancey:
Any plan on moving forward with these changes?

s-yadav:
Closing this PR as in v5 it can be done outside of the lib. Added an example here.
https://codesandbox.io/s/parentheses-for-negation-forked-jn42cp

Also, update the customization doc to list this: https://s-yadav.github.io/react-number-format/docs/customization/#using-parentheses-to-express-negative-numbers

PR #302, opened 2019-04-09
Add a boolean prop to allow leading zeros

nmalzieu:
Hi,

this is a small proposition for a prop to disable the removal of leading zeros on `onBlur` method.

This is because of a special case where I want the user to be able to enter a number with leading zeros that will become the suffix of another number, hence the auto removal of leading zeros breaks the behaviour.

Don't hesitate if you have any question !

Thanks

### DIFF
--- a/test/library/input_numeric_format.spec.js
+++ b/test/library/input_numeric_format.spec.js
@@ -365,7 +365,7 @@ describe('Test NumberFormat as input with numeric format options', () => {
     expect(wrapper.state().value).toEqual('$20,000.25');
   });
 
-  it('should remove leading 0s while user go out of focus', () => {
+  it('should remove leading 0s while user go out of focus and allowLeadingZeros is false', () => {
     const wrapper = shallow(<NumberFormat value={23456.78} thousandSeparator={','} decimalSeparator={'.'} prefix={'$'}/> );
 
     simulateKeyInput(wrapper.find('input'), '0', 1);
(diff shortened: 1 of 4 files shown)


nmalzieu:
Hello,

just checking what you think about this PR and if it could be merge. Would prevent me from using a fork in my project :-)

Thanks a lot!

s-yadav:
Sorry for the late reply, Been busy on other things lately. The PR looks good. 
The leading zero should not have a thousand separators though.

test/library/input_numeric_format.spec.js
```diff
@@ -381,6 +381,22 @@ describe('Test NumberFormat as input with numeric format options', () => {
...
+    simulateBlurEvent(wrapper.find('input'));
+
+    expect(wrapper.state().value).toEqual('$023,456.78');
+
+    wrapper.setProps({value: 10000.25});
+    wrapper.update();
+
+    simulateKeyInput(wrapper.find('input'), 'Backspace', 2);
+    simulateBlurEvent(wrapper.find('input'));
+    expect(wrapper.state().value).toEqual('$0,000.25');
```
s-yadav: It's a wrong test. The leading 0 should be having a thousand separators. I remember fixing this issue. Check after new code pull if this persists?

Sonia-L: When will the new feature to allow leading zeros be available?

techno1ology: I am also be interested in this feature

techno1ology: Hi @s-yadav.  I forked this project to complete this feature.  With latest the above test is still passing.  What result are you expecting?  are you expecting '$0000.25' instead of '$0,000.25'

s-yadav: Yes. It should be $0000.25

techno1ology:
Resolved by https://github.com/s-yadav/react-number-format/pull/331

s-yadav:
Released this on v4.2.0

PR #318, opened 2019-06-01
formatAsANumber ignores decimalSeparator property

szogun1987:
It always tries to find dot.
I haven't tested this solution. It is based on static code analysis exclusively


szogun1987:
Sorry I found the wrong point. For some reason, your library has a problem with space as thousand a separator and coma as a decimal one

PR #325, opened 2019-07-12
resolves #293 - Calls clearTimeOut in onBlur function

### LINKED ISSUE #293: onFocus={event => event.target.select()} introduces infinite loop bug
If you click on a NumberFormat component that has the property onFocus={event => event.target.select()} to set focus on its contents, then click out of the browser (on another screen, or in the developer console, etc.), then click another NumberFormat component that also has onFocus={event => event.target.select()}, an infinite loop occurs where both inputs constantly set themselves as the focus.

Test an example of this with this [Pen](https://codepen.io/anon/pen/xBYNVW).

Is there currently a way to work around this issue? My use case requires my formatted inputs to have this onFocus function.

edwardkads:
https://github.com/s-yadav/react-number-format/issues/293#issuecomment-510965743

### DIFF
--- a/src/number_format.js
+++ b/src/number_format.js
@@ -90,6 +90,7 @@ class NumberFormat extends React.Component {
   onMouseUp: Function
   onFocus: Function
   onBlur: Function
+  focusTimeout: number
   focusedElm: HTMLElement
   selectionBeforeInput: {
     selectionStart: number,
@@ -731,8 +732,11 @@ class NumberFormat extends React.Component {
     const {format, onBlur} = props;
     let {numAsString} = state;
     const lastValue = state.value;
-
     this.focusedElm = null;
+    if (this.focusTimeout){
+      clearTimeout(this.focusTimeout);
+    }
+
 
     if (!format) {
       numAsString = fixLeadingZero(numAsString);
@@ -852,8 +856,7 @@ class NumberFormat extends React.Component {
     e.persist();
 
     this.focusedElm = e.target;
-
-    setTimeout(() => {
+    this.focusTimeout = setTimeout(() => {
       const el = e.target;
       const {selectionStart, selectionEnd, value = ''} = el;
 


guipra:
@s-yadav, any reason this can't be merged yet? Is there a timeline for the next release? Is there something we can do to help? Thanks!

s-yadav:
@edwardkads @guipra. Sorry I missed this earlier. The changes look good. @guipra, @edwardkads Can you add a failing test case for it. So we don't end up on the regression issue later. 

Let me know if you need any help in writing the test case.
Thanks!

guipra:
@s-yadav, I think this is it, but I'm not sure in which module should this test be added. Lemme know what you think:

```
it('should not call onFocus if blur event is dispatched in the same event loop', () => {
  const onFocusSpy = jasmine.createSpy(() => null);
  const wrapper = mount(<NumberFormat onFocus={onFocusSpy} />);
  const input = wrapper.find('input');
  simulateFocusEvent(input);
  simulateBlurEvent(input);
  expect(onFocusSpy).not.toHaveBeenCalled();
});
```

guipra:
I think I got it right now:

```js
it('should not call onFocus if blur event is dispatched in the same event loop', done => {
  const onFocusSpy = jasmine.createSpy(() => null);

  const wrapper = mount(<NumberFormat onFocus={onFocusSpy} />);
  const input = wrapper.find('input');

  simulateFocusEvent(input);
  simulateBlurEvent(input);

  setTimeout(() => {
    expect(onFocusSpy).not.toHaveBeenCalled();
    done();
  }, 0);
});
```

Still wondering in which test module this would fit in. I couldn't find any other tests related to the `onFocus` prop to put it there.

s-yadav:
@guipra The test looks good. Will merge both the PR. Thanks.

guipra:
Thanks, @s-yadav. Do you know when we could expect this to land in the next release?

s-yadav:
@guipra Release now. It's available in v4.3.1.

PR #327, opened 2019-07-25
Format function should support decimal

wapgear:
On our project we had a problem with your library - for some reason with `typeof props.format === "function"`  library worked only with integers, with this PR it'll also support decimal values.


wapgear:
@s-yadav hey sorry if I disturb you but what do you think about this?

src/number_format.js
```diff
@@ -417,15 +417,13 @@ class NumberFormat extends React.Component {
...
-      val = this.removePrefixAndSuffix(val);
-      val = this.getFloatString(val);
-    } else if (typeof format === 'string') {
+    if (typeof format === 'string') {
       val = this.removePatternFormatting(val);
     } else if (typeof removeFormatting === 'function') { //condition need to be handled if format method is provide,
       val = removeFormatting(val);
     } else {
-      val = (val.match(/\d/g) || []).join('')
+      val = this.removePrefixAndSuffix(val);
```
s-yadav: We should not handle other formatting props, when format function is provided.

s-yadav:
I don't think we should handle prefix / suffix, and other formatting props if format prop is defined as function. If we do so that will lead to lots of corner case which is not meant to be solved by here.

We can look for some regex instead which allows all the valid character of a number and let the format method handle the cases.

Also, If a user needs to have more control on removing formatting he can pass removeFormat method. 
I guess we can work on improving the documentation of format (as function) and removeFormatting method, so people are aware of it while using custom format function. What do you say?

PR #331, opened 2019-08-06
Add allowLeadingZeros property.  Bypass applying thousandSeparator when left side of decimal is zero

techno1ology:
@s-yadav I reviewed the code regarding not applying the thousandSeparator when the number is zero.  It din't appear to exist so I added it for you. Thanks!

### DIFF
--- a/src/number_format.js
+++ b/src/number_format.js
@@ -47,6 +47,7 @@ const propTypes = {
   customInput: PropTypes.elementType,
   allowNegative: PropTypes.bool,
   allowEmptyFormatting: PropTypes.bool,
+  allowLeadingZeros: PropTypes.bool,
   onValueChange: PropTypes.func,
   onKeyDown: PropTypes.func,
   onMouseUp: PropTypes.func,
@@ -68,6 +69,7 @@ const defaultProps = {
   suffix: '',
   allowNegative: true,
   allowEmptyFormatting: false,
+  allowLeadingZeros: false,
   isNumericString: false,
   type: 'text',
   onValueChange: noop,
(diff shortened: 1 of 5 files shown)


techno1ology:
Hi @s-yadav. I'm concerned with the lib files.  I used compile to generate, but there are some unwanted changes.  I'm thinking maybe I run bundle instead? It looks like I need to install yarn to do this?  I will try later.  If you know what I need to do, please let me know.  Oh and I found you commit where you stated that you fixed the thousandSeparator with leading zeroes, but it didn't appear to work. Thanks!

src/number_format.js
```diff
@@ -465,7 +467,9 @@ class NumberFormat extends React.Component {
     if (decimalScale !== undefined) afterDecimal = limitToScale(afterDecimal, decimalScale, fixedDecimalScale);
 
     if(thousandSeparator) {
-      beforeDecimal = applyThousandSeparator(beforeDecimal, thousandSeparator, thousandsGroupStyle);
+      const beforeDecimalNumAsString = this.removeFormatting(beforeDecimal);
+      const beforeDecimalFloatValue = parseFloat(beforeDecimalNumAsString);
```
s-yadav: I don't think these changes are required. 
The bug is here.
https://github.com/s-yadav/react-number-format/blob/master/src/utils.js#L29
The logic works for cases like `00001,234` but doesn't work only with 0s. We should add check for index being -1. 
Changing
```
const index = str.search(/[1-9]/);
```
to 
```
let index = str.search(/[1-9]/);
index = index === -1 ? str.length : index;
```
on util will fix the bug

src/number_format.js
```diff
@@ -728,14 +732,14 @@ class NumberFormat extends React.Component {
...
-    const {format, onBlur} = props;
+    const {format, onBlur, allowLeadingZeros} = props;
     let {numAsString} = state;
     const lastValue = state.value;
 
     this.focusedElm = null;
 
     if (!format) {
-      numAsString = fixLeadingZero(numAsString);
+      numAsString = allowLeadingZeros ? numAsString : fixLeadingZero(numAsString);
```
s-yadav: Change it to 
```js
if (!allowLeadingZeros) {
   numAsString = fixLeadingZero(numAsString);
}
```
This has less logical complexity with if is 1, while with ternary it's 2 (ternary is if + else). 
Can be on the same line as well.

src/number_format.js
techno1ology: great feedback.  updated

src/number_format.js
techno1ology: updated

PR #339, opened 2019-09-05
Added onBlur event type for fixed scale.

erdembas:
I think it will be very useful


berkayakcay:
Ohh! Amazing! That what we need!

baturox:
It will be good for project

PR #343, opened 2019-09-13
Add input name to onValueChange return object

dejorrit:
Add a `name` (if set via props) to the object returned from the `onValueChange` method. This helps writing a clean generic callback method to update state.

**Example usage**

```jsx
const MyComponent = () => {
    const [values, setValues] = useState({});

    function handleOnChange({ floatValue, name }) {
        setValues({
            ...values,
            [name]: floatValue
        });
    }

    return (
        <div>
            <NumberFormat name="anInput" onValueChange={handleOnChange} />
            <NumberFormat name="anotherInput" onValueChange={handleOnChange} />
            <NumberFormat name="moreInput" onValueChange={handleOnChange} />
        </div>
    )
}
```


kolharsam (commented):
This LGTM.

cc @s-yadav

kolharsam:
@dejorrit Could you resolve the conflict on this PR? 
Thanks! :)

s-yadav:
I guess the better way for this particular case would be binding the name with handleOnChange on source side.

```js
<NumberFormat name="anInput" onValueChange={handleOnChange.bind(this, 'anInput')} />
```

Moreover, there is another suggestion going on where we can pass source object as second parameter.
https://github.com/s-yadav/react-number-format/issues/429#issuecomment-731706973

dejorrit:
> @dejorrit Could you resolve the conflict on this PR?
> Thanks! :)

Done!

kolharsam:
@s-yadav 

> I guess the better way for this particular case would be binding the name with handleOnChange on source side.

This suggestion does make sense to me too. 

And about your comment [here](https://github.com/s-yadav/react-number-format/issues/429#issuecomment-731706973), I think this would nice enhancement on the `onValueChange` method

Issue #324, opened 2019-07-08
Comma character could be taken as period(default decimalSeparator)


DeylEnergy:
Hello. There is a little problem that took me a long time. How to make so that the component treats ',' as '.'? Tried to manually invoke NumberFormat.prototype.onKeyDown from onKeyDown prop including SyntheticEvent with key replaced to '.' 

Is there an escape hatch for it? Will appreciate any suggestions.

duncanawerbuck:
@DeylEnergy, have you tried setting the `thousandSeparator` and `decimalSeparator` props?

They're documented [here](https://github.com/s-yadav/react-number-format#props).

DeylEnergy:
Thank you @duncanawerbuck I already tried to change things around those props. The thing is I try to set ```decimalSeparator``` to ```,``` form could take both ```.``` and ```,```. However, I want ```,``` to work  as  ```.``` set to ```decimalSeparator```. Even tried to literally change the ```key``` property of ```event``` to ```.``` when ```onKeyDown``` takes ```,``` as ```key``` property

duncanawerbuck:
> make so that the component treats ',' as '.'

I'm not sure I understand. The component considers two things:
* Which character you would like to be the thousand separator. If you use US/UK number formatting, you would specify `,` as the `decimalSeparator` prop. Note that this isn't something that the user types. You also dont' have to set it at all if you don't want a thousand separator to appear. If you have specified it, the component will automatically insert it when a certain number of digits are entered.
* Which character the user will type to indicate they are about to type decimal values.

So assuming US/UK formatting, consider a user who might wish to enter "one thousand dollars and fifty cents" (they want to end up seeing `1,000.50`).

Step 1: They type `1` (note they cannot type `,`)
Step 2: They type `0` (they see `10` - no formatting has been applied)
Step 3: They type `0` (they see `100` - no formatting has been applied)
Step 4: They type `0` (they see `1,000` - the thousand separator, i.e. comma, is automatically displayed)
Step 5: They now wish to add the fifty cents, so they press `.`  (they see `1,000.`). The value at this point is "one thousand".
Step 6: They type `5` (they see `1,000.5`)
Step 7: They type `0` (they see `1,000.50`)

[...]

Issue #261, opened 2018-12-19
Custom decimal separator is not working (european style)


Lysander:
When I define the component like this I don't get decimal separator to work at all. It is invisible, gone after typing:

```
      return (
        <NumberFormat
          {...other}
          getInputRef={inputRef}
          onValueChange={values => {
            onChange({
              target: {
                value: values.value
              }
            });
          }}
          allowNegative={false}
          decimalSeparator={","}
          decimalPrecision={2}
          thousandSeparator={"."}
          suffix={"€"}
        />
      );
```

Also I got this warning
React does not recognize the decimalPrecision prop on a DOM element.

But the main thing is that decimal separator does not work.

I want to be able to enter 12,25€ value but at the moment it is formatted to 1.225€

_Originally posted by @atipi in https://github.com/s-yadav/react-number-format/issues/191#issuecomment-448151521_

You can try yourself: https://codesandbox.io/s/5x490556vk

Just enter a value and you will get an error!

    The above error occurred in the <NumberFormat> component:
        in NumberFormat (created by NumberFormatCustom)
        in NumberFormatCustom (created by InputBase)
        in div (created by InputBase)
        in InputBase (created by Context.Consumer)
        in WithFormControlContext(InputBase) (created by WithStyles(WithFormControlContext(InputBase)))
        in WithStyles(WithFormControlContext(InputBase)) (created by Input)
        in Input (created by WithStyles(Input))
        in WithStyles(Input) (created by TextField)
        in div (created by FormControl)
        in FormControl (created by WithStyles(FormControl))
        in WithStyles(FormControl) (created by TextField)
        in TextField (created by FormattedInputs)
        in div (created by FormattedInputs)
        in FormattedInputs (created by WithStyles(FormattedInputs))
        in WithStyles(FormattedInputs)

    Consider adding an error boundary to your tree to customize error handling behavior.

    Visit https://fb.me/react-error-boundaries to learn more about error boundaries.

apuntovanini:
Bug is confirmed, just can't use the "," as decimal separator

s-yadav:
As mentioned in doc
https://github.com/s-yadav/react-number-format#values-object
and Notes (6th point)
https://github.com/s-yadav/react-number-format#notes-and-quirks
values.value will be numeric string, if you are passing numeric string as value, you should set isNumericString={true} on the number format, so module can identify it's a formatted value or numeric string.

Here is the modified example of NumberFormat with Material UI for following case
[...]

Issue #223, opened 2018-09-17
Multiple custom decimalSeparator


DylanMerigaud:
Hi,
I would like to be able to type multiple characters to produce the same decimalSeparator.
For example '.' or ',' and should produce ','.
This is common for french to be able to type a dot or a comma since our decimal separator is a comma.

Any help is welcomed :)

DylanMerigaud:
[Existing issue](https://github.com/s-yadav/react-number-format/issues/115#issuecomment-342141253)

Issue #133, opened 2017-12-10
Any way to autoreplace ',' with '.' on typing?


drakmail:
I want to implement autoreplace ',' with '.' on typing in input. Does there any ways to did it with react-number-format?

s-yadav:
Try this 
https://github.com/s-yadav/react-number-format/issues/115#issuecomment-342141253
However, this is not supported on android chrome. So you can add check if keyCode is not supported for any browser don't do this.

drakmail:
Thank you, I'll try it

PR #345, opened 2019-09-24
Add `allowedDecimalSeparators` prop

macobo:
When defined, this prop defines which keys to accept as decimal
separators. This is useful in cases where users might want to enter both
comma and dot as separators

Related issues: #324, #133, #223, #261

### DIFF
--- a/src/number_format.js
+++ b/src/number_format.js
@@ -23,6 +23,7 @@ import {
 const propTypes = {
   thousandSeparator: PropTypes.oneOfType([PropTypes.string, PropTypes.oneOf([true])]),
   decimalSeparator: PropTypes.string,
+  allowedDecimalSeparators: PropTypes.arrayOf(PropTypes.string),
   thousandsGroupStyle: PropTypes.oneOf(['thousand', 'lakh', 'wan']),
   decimalScale: PropTypes.number,
   fixedDecimalScale: PropTypes.bool,
@@ -197,15 +198,22 @@ class NumberFormat extends React.Component {
 
   getSeparators() {
     const {decimalSeparator} = this.props;
-    let {thousandSeparator} = this.props;
+    let {thousandSeparator, allowedDecimalSeparators} = this.props;
 
     if (thousandSeparator === true) {
       thousandSeparator = ','
     }
+    if (!allowedDecimalSeparators) {
+      allowedDecimalSeparators = [decimalSeparator]
+      if (decimalSeparator !== '.') {
+        allowedDecimalSeparators = allowedDecimalSeparators.concat(['.'])
+      }
+    }
 
     return {
       decimalSeparator,
-      thousandSeparator
+      thousandSeparator,
+      allowedDecimalSeparators,
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/number_format.js
```diff
@@ -603,13 +611,13 @@ class NumberFormat extends React.Component {
...
-    const {decimalSeparator} = this.getSeparators();
+    const {allowedDecimalSeparators, decimalSeparator} = this.getSeparators();
     const lastNumStr = this.state.numAsString || '';
     const {selectionStart, selectionEnd} = this.selectionBeforeInput;
     const {start, end} = findChangedIndex(lastValue, value);
 
-    /** Check if only . is added in the numeric format and replace it with decimal separator */
-    if (!format && decimalSeparator !== '.' && start === end && value[selectionStart] === '.') {
+    /** Check for any allowed decimal separator is added in the numeric format and replace it with decimal separator */
+    if (!format && start === end && allowedDecimalSeparators.includes(value[selectionStart])) {
```
macobo: Note: This might not work in all browsers. Should I extract a util instead?

macobo:
Hi @s-yadav, mind taking a look? Happy to iterate on this as/if needed.

src/number_format.js
s-yadav: You can use .indexOf instead.

src/number_format.js
```diff
@@ -197,15 +198,22 @@ class NumberFormat extends React.Component {
...
     const {decimalSeparator} = this.props;
-    let {thousandSeparator} = this.props;
+    let {thousandSeparator, allowedDecimalSeparators} = this.props;
 
     if (thousandSeparator === true) {
       thousandSeparator = ','
     }
+    if (!allowedDecimalSeparators) {
+      allowedDecimalSeparators = [decimalSeparator]
+      if (decimalSeparator !== '.') {
```
s-yadav: No need for this if condition.  You can just define 
```js
allowedDecimalSeparators = ['.', decimalSeparator]
```
It's okay if dot gets duplicate as we are using this array just for checking includes.

macobo:
Thanks for the feedback @s-yadav. I've updated the PR.

s-yadav:
Can you also update the readme for the new props

macobo:
Also done @s-yadav.

s-yadav:
I see that you have updated the example file, we will need it to update on README file.

macobo:
Oops - sorry, tried to push to the wrong remote thus the PR didn't get updated. Done.

s-yadav:
Merged. Right now I am traveling. Will test it out and release it on npm tomorrow.

macobo:
Thank you!

tenkij:
Thank you!
Little problems:
https://codesandbox.io/s/react-number-format-demo-9y7j7

1) If you paste a number '11,11', then the formatted number will be '1 111'. If you enter a number '11.11', then the formatted number will be '11.11'.

2) Press the “11, 11” button. The input value will be “11, 11”. Then click on the input. After losing focus,  the input value will be “1 111”.

macobo:
Hi @tenkij it's worth creating a new issue for these rather than on a merged PR. The pasting issue seems to exist before this PR at least.

s-yadav:
@tenkij the allowedDecimalSeparators is to only allow user to type a character which will be converted to passed decimalSeparator prop. When you paste or change a state value, the incorrect characters are removed. 
But when you pass isNumericString true, it is an indication to NumberFormat that the passed value prop is a valid numeric string (without formatting) do not try to process it.

s-yadav:
But there seem to be some inconsistencies on paste, which I will look into.

tenkij:
@macobo, ok.
Problem only if the following conditions are true:
1) allowedDecimalSeparators={[",", "."]}
2) decimalSeparator = "." 
3) paste '11,11'

paste '11.11' - ok.

PR #347, opened 2019-09-26
Allow null value in TypeScript typing

erikpuk:
Pretty trivial PR, but since you need to be able to set the `value` prop to `null`, seems like it should be allowed in the typings.

That said, the behavior itself did confuse me. [This line specifically](https://github.com/s-yadav/react-number-format/blob/master/src/number_format.js#L142), where you ignore the `value` prop if it is `undefined`....  I guess you are using `value={undefined}` as a signal from the consumer that they want the NumberFormat to be stateful?

Might be better to have an explicit `stateless` prop. But maybe there's just some inevitable confusion when you can use the same component in either way.

I could add a bit to the readme about passing `null` if you are trying to use the NumberFormat as a stateless component.

### DIFF
--- a/typings/number_format.d.ts
+++ b/typings/number_format.d.ts
@@ -39,7 +39,7 @@ declare module "react-number-format" {
     format?: string | FormatInputValueFunction;
     removeFormatting?: (formattedValue: string) => string;
     mask?: string | string[];
-    value?: number | string;
+    value?: number | string | null;
     defaultValue?: number | string;
     isNumericString?: boolean;
     customInput?: React.ComponentType<any>;


s-yadav:
Nice catch.

> I guess you are using value={undefined} as a signal from the consumer that they want the NumberFormat to be stateful

It's more of when value prop is not passed at all. Though `value={undefined}` will work. As `null` as a value also works for making an uncontrolled component in native inputs. We should update our codebase to add null check as well. 

@erikpuk It would be great if you can update this PR to have null check as well.

sarnendumukherjee:
@s-yadav Right now is there a way to set value as null. because setting the value as null is not permitted as per the typings but that works.

s-yadav:
@sarnendumukherjee I think passing null will break. https://github.com/s-yadav/react-number-format/blob/master/src/number_format.js#L152

We should check for null on this line as well. Would you like to give a PR for this?

sarnendumukherjee:
HI @s-yadav , I created a small snip with an example of what am i trying to do and how it works but the TS will throw error and will not build as null is not a passable value for the value prop.
https://codesandbox.io/s/musing-haze-ddkuv?file=/src/App.js
You can change the initial value and try it out. 
My worry is if we add the null check there this reset feature will stop working. 
I was looking for a way to reset the value and setting the value as undefined does not reset the field.

s-yadav (approved):

PR #355, opened 2019-10-03
getInputRef should not be default noop

mvklingeren:
If getInputRef is noop by default, a ref is given to the customInput, when used. 
Causing a ugly warning in React.


s-yadav:
PR looks good. Thanks.

PR #356, opened 2019-10-03
Suggest to use Yarn for dependencies

bokub:
Because there is a `yarn.lock` and no `package-lock.json`, it's better for developers to use Yarn as a dependency management tool.

I've updated the README.

Also, I didn't know if I should change the `npm` commands too, for example
- `npm start` > `yarn start`
- `npm run bundle` > `yarn bundle`

What's your opinion on this ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,8 +286,9 @@ Feature Addition
 
 
 ### Development
-- Download the zip
-- `npm install`
+- Clone the repository or download the zip
+- `npm i -g yarn` to download Yarn
+- `yarn` to install dependencies
 - `npm start` to run example server
 - `npm run test` to test changes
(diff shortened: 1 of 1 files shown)


s-yadav:
Agree. Will merge. Thanks.

README.md
```diff
@@ -286,8 +286,9 @@ Feature Addition
 
 
 ### Development
-- Download the zip
-- `npm install`
+- Clone the repository or download the zip
+- `npm i -g yarn` to download Yarn
+- `yarn` to install dependencies
 - `npm start` to run example server
 - `npm run test` to test changes
```
s-yadav: We can also update other commands for yarn.

bokub:
Done !

PR #364, opened 2019-10-14
Reflect possible undefined floatValue in typings

LinusU:
This value can be set explicitly to `undefined`, e.g. when the user clears the field:

https://github.com/s-yadav/react-number-format/blob/b233bd68c7cd16c63287c3aa4df576c2ff80eca9/src/number_format.js#L225


LinusU:
ping @s-yadav ☺️

PR #370, opened 2019-11-12
Notify value changes ASAP to prevent infinite setState loop

zhusee2:
Patch `<NumberFormat>` to notify parent components about value changes as soon as possible, to prevent infinite `setState` loop as discovered in https://github.com/s-yadav/react-number-format/issues/277#issuecomment-509938027

That way when `<NumberFormat>` is being controlled, its parent component can learn about new value at the first time and then trigger a render with new value passed to `<NumberFormat>` prop.

### DIFF
--- a/src/number_format.js
+++ b/src/number_format.js
@@ -697,10 +697,9 @@ class NumberFormat extends React.Component {
 
     //update state if value is changed
     if (formattedValue !== lastValue) {
-      this.setState({value : formattedValue, numAsString}, () => {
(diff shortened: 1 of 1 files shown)


s-yadav:
@zhusee2 Can we have a failing testcase with this.?

s-yadav:
Thanks for the fix. 
This been added in 4.4.1

zhusee2:
Thanks for accepting this patch!
Sorry I didn't have time to create a failing test before.

PR #376, opened 2019-11-30
Fixed issue with dot separator when decimalScale is set to zero #375

### LINKED ISSUE #375: Incorrect dot handler for input with decimalScale=0
Hi, thank you for your library. It looks cool.

I found the problem when I created the component with the property `decimalScale`.
Steps to reproduce:
1. Create the NumberFormat component with next props:
```
<NumberFormat 
     value={4200} 
     thousandSeparator={true}
     decimalScale={0}
/>
```
2. Try to press or paste the **dot** symbol between zeros
3. Last zero symbol was removed 

You can see my example [here](https://codepen.io/sergei97k/pen/YzPKELM)

P.S. I resolved this issue. I added the property `decimalSeparator={null}`. But I'm not sure, that it's the clear solution (because decimalSeparator should get only "string" type).

sergei97k:
Fixes #375

### DIFF
--- a/src/number_format.js
+++ b/src/number_format.js
@@ -615,7 +615,8 @@ class NumberFormat extends React.Component {
 
     /** Check for any allowed decimal separator is added in the numeric format and replace it with decimal separator */
     if (!format && start === end && allowedDecimalSeparators.indexOf(value[selectionStart]) !== -1  ) {
-      return value.substr(0, selectionStart) + decimalSeparator + value.substr(selectionStart + 1, value.length);
+      const separator = this.props.decimalScale === 0 ? '' : decimalSeparator;
+      return value.substr(0, selectionStart) + separator + value.substr(selectionStart + 1, value.length);
     }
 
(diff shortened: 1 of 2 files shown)


src/number_format.js
```diff
@@ -615,7 +615,8 @@ class NumberFormat extends React.Component {
 
     /** Check for any allowed decimal separator is added in the numeric format and replace it with decimal separator */
     if (!format && start === end && allowedDecimalSeparators.indexOf(value[selectionStart]) !== -1  ) {
-      return value.substr(0, selectionStart) + decimalSeparator + value.substr(selectionStart + 1, value.length);
+      const separator = this.props.decimalScale === 0 ? '' : decimalSeparator;
```
s-yadav: Destruct decimalScale on top of this function.

s-yadav:
Thanks, the changes look good. And sorry for the very late reply. 
There is just a small change. I will merge this and make that change.

PR #389, opened 2020-02-23
fix: Clear focusTimeout timeout on unmount

doniyor2109:
`onFocus` is handled asynchronously with setTimeout. When testing inputs with react-number-format, it prints a warning saying it is updating the unmounted component.

Clear focusTimeout on unmount.

### DIFF
--- a/src/number_format.js
+++ b/src/number_format.js
@@ -134,10 +134,7 @@ class NumberFormat extends React.Component {
   }
 
   componentWillUnmount() {
-    if (this.focusTimeout != null) {
-      clearTimeout(this.focusTimeout);
-      this.focusTimeout = undefined;
-    }
+    clearTimeout(this.focusTimeout);
   }
 
   updateValueIfRequired(prevProps: Object) {
@@ -760,10 +757,8 @@ class NumberFormat extends React.Component {
     let {numAsString} = state;
     const lastValue = state.value;
     this.focusedElm = null;
-    if (this.focusTimeout){
-      clearTimeout(this.focusTimeout);
-      this.focusTimeout = undefined;
-    }
+
+    clearTimeout(this.focusTimeout);
 
 
     if (!format) {
--- a/src/number_format.js
+++ b/src/number_format.js
@@ -133,6 +133,12 @@ class NumberFormat extends React.Component {
     this.updateValueIfRequired(prevProps);
   }
 
+  componentWillUnmount() {
+    if (this.focusTimeout != null) {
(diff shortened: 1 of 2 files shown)


src/number_format.js
```diff
@@ -133,6 +133,13 @@ class NumberFormat extends React.Component {
     this.updateValueIfRequired(prevProps);
   }
 
+  componentWillUnmount() {
+    if (this.focusTimeout != null) {
+      clearTimeout(this.focusTimeout);
+      this.focusTimeout = undefined;
```
s-yadav: We don't have to set it to undefined. only clearTimeout is enough.

src/number_format.js
```diff
@@ -133,6 +133,13 @@ class NumberFormat extends React.Component {
     this.updateValueIfRequired(prevProps);
   }
 
+  componentWillUnmount() {
+    if (this.focusTimeout != null) {
```
s-yadav: We don't need this.focusTimeout check. clearTimeout early bails out if we call it with undefined or non-active timeouts.
Same thing we can do it for onBlur.

src/number_format.js
```diff
@@ -755,6 +762,7 @@ class NumberFormat extends React.Component {
     this.focusedElm = null;
     if (this.focusTimeout){
```
s-yadav: ^^ We can remove the if condition.

test/library/keypress_and_caret.spec.js
```diff
@@ -342,6 +342,24 @@ describe('Test keypress and caret position changes', () => {
...
+      const instance = wrapper.instance();
+
+      simulateFocusEvent(wrapper.find('input'), 0, 0, setSelectionRange);
+
+      expect(instance.focusTimeout).toBeTruthy();
+
+      wrapper.unmount();
+      jasmine.clock().tick(1);
+
+      expect(instance.focusTimeout).toBeFalsy();
```
s-yadav: Instead of checking if focusTimeout is empty let's check if the function inside the timeout is called. We can use onFocus prop to check this.

test/library/keypress_and_caret.spec.js
```diff
@@ -342,6 +342,24 @@ describe('Test keypress and caret position changes', () => {
...
+
+      jasmine.clock().uninstall();
```
s-yadav: If the spec fails, other test will fail, as the clock will still be mocked by this spec. 
You should clear mock on afterEach. Do unInstall there.

PR #409, opened 2020-05-04
Ios fixes

s-yadav:


dist/react-number-format.es.js
```diff
@@ -346,6 +346,9 @@ function getCurrentCaretPosition(el) {
   /*Max of selectionStart and selectionEnd is taken for the patch of pixel and other mobile device caret bug*/
   return Math.max(el.selectionStart, el.selectionEnd);
 }
+function addInputMode(format) {
+  return format || !(navigator.platform && /iPhone|iPod/.test(navigator.platform));
```
wojtekmaj, line 350: This misses match on iPad :)

s-yadav: This is intentional. Ipad shows a correct keyboard on `inputmode=numeric` which have `,` and `.`. I see there is a typo on the test case statement. Will fix that.

PR #444, opened 2020-10-17
Handle scientific notation numbers

yhnavein:
added unit tests to prove it does the work
chore: added vscode config that won't break anything in this project

The reason I did this change is that before when I was formatting small numbers I was seeing huge numbers instead. Take a look:
```
<NumberFormat displayType="text" value={0.00000041} thousandSeparator={' '} decimalScale={20} />
or
<NumberFormat displayType="text" value={4.1e-7} thousandSeparator={' '} decimalScale={20} />
```
would render `5 000 000.000` for some strange reason. Apparently numbers given in a scientific notation are handled very poorly.

I've added a simple support for scientific notation to the NumberFormat component and now it looks much better.

BTW - I've added very simple VSCode config so that it won't prettify whole files on save.


yhnavein:
Bumping topic...

lorenzoa7:
Any updates on this?

PR #462, opened 2020-11-22
Incremental PR towards #347

kolharsam:
This PR is an incremental PR against #347 based on this [comment](https://github.com/s-yadav/react-number-format/pull/347#issuecomment-683865867)

### DIFF
--- a/test/library/input.spec.js
+++ b/test/library/input.spec.js
@@ -11,7 +11,6 @@ import {
   shallow,
   mount
 } from '../test_util';
-import { exact } from 'prop-types';
 
 /*** format_number input as input ****/
 describe('NumberFormat as input', () => {
--- a/test/library/input.spec.js
+++ b/test/library/input.spec.js
@@ -11,6 +11,7 @@ import {
   shallow,
   mount
 } from '../test_util';
+import { exact } from 'prop-types';
 
 /*** format_number input as input ****/
 describe('NumberFormat as input', () => {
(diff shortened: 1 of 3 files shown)


kolharsam:
@s-yadav I'm not sure if a changelog entry is needed for this PR. I could just add one for #347 directly.

s-yadav:
@kolharsam Can you add failing test case, which this PR fixes.

kolharsam:
@s-yadav I have added some tests with regards to these changes but I failed to think of a failing test. Would passing an empty string be considered a failing test? Would appreciate some input here. If so, then we could change the condition to just be `!props.value ? .... : .....` [here](https://github.com/s-yadav/react-number-format/pull/462/files#diff-a1be15e225934962f6ce9ddb7475f5945550187846ed4cd69c3d0a51f849abbcR152)

s-yadav (approved):

s-yadav:
Looks good. Approved.

s-yadav:
@kolharsam Looks like some specs are failing after this merge. Can you check that.

kolharsam:
@s-yadav I think the specs are failing because the original PR #347 hasn't been merged yet and these changes were increments over the mentioned PR. So yeah, I guess things would get fixed if we merge #347.

s-yadav:
Merged. Working.

test/library/input.spec.js
```diff
@@ -56,6 +56,33 @@ describe('NumberFormat as input', () => {
     expect(wrapper.find('input').instance().value).toEqual('$2,456,981');
   });
 
+  it('should load the default value when initial value is null', () => {
+    const wrapper = mount(<NumberFormat value={null} defaultValue={89} />);
+    expect(wrapper.state().value).toEqual('89');
```
s-yadav, line 61: We should never check for state value. As it's implementation details. Instead we should test for the end side-effect.
```js
expect(wrapper.find('input').instance().value).toEqual('89');
```

test/library/input.spec.js
```diff
@@ -56,6 +56,33 @@ describe('NumberFormat as input', () => {
...
+  });
+
+  it('should load the prevous valid value if the state is changed to null', () => {
```
s-yadav, line 64: Typo on the statement.
```
  should load the previous valid value
```

test/library/input.spec.js
```diff
@@ -56,6 +56,33 @@ describe('NumberFormat as input', () => {
...
+        return (<NumberFormat value={this.state.testState} />)
+      }
+    }
+
+    const wrapper = mount(<WrapperComponent />);
+    const input = wrapper.find('input');
+    const domInput = input.instance();
+
+    expect(domInput.value).toEqual('90');
+    wrapper.setState({testState: null});
```
s-yadav, line 82: This spec could be simplified you just mount `<NumberFormat value={this.state.testState} />` and then change the props through setProps.

Also FYI. Try avoiding interacting with states in specs. You shouldn't depend much on implementation detail.

s-yadav:
@kolharsam The test is still failing when I am trying to run the test.
Screenshot 2020-12-18 at 12 18 07 PM

Also, missed reviewing specs earlier. Added some comments there.

s-yadav:
@kolharsam Can you try pulling code from origin and rerun the test.

kolharsam:
Will be making a new PR to fix this ASAP

Issue #446, opened 2020-10-21
NumberFormatProps defaultValue type doesn't match @types/react


benneq:
When using `<NumberFormat />` with react material `<TextField />` I get:

```
Type 'FC<NumberFormatProps>' is not assignable to type '"abbr" | "address" | "article" | "aside" | "b" | "bdi" | "bdo" | "big" | "blockquote" | "caption" | "cite" | "code" | "data" | "dd" | "del" | "details" | "dfn" | "dt" | "em" | "figcaption" | ... 39 more ... | undefined'.
  Type 'FunctionComponent<NumberFormatProps>' is not assignable to type 'FunctionComponent<InputBaseComponentProps>'.
    Types of parameters 'props' and 'props' are incompatible.
      Type 'PropsWithChildren<InputBaseComponentProps>' is not assignable to type 'PropsWithChildren<NumberFormatProps>'.
        Type 'PropsWithChildren<InputBaseComponentProps>' is not assignable to type 'NumberFormatProps'.
          Types of property 'defaultValue' are incompatible.
            Type 'string | number | readonly string[] | undefined' is not assignable to type 'string | number | undefined'.
              Type 'readonly string[]' is not assignable to type 'string | number | undefined'.
                Type 'readonly string[]' is not assignable to type 'string'.ts(2322)
```

The source of this issue comes from `@types/react`:
```
interface HTMLAttributes<T> extends AriaAttributes, DOMAttributes<T> {
    ...
    defaultValue?: string | number | ReadonlyArray<string>;
    ...
}
```

Though I suggest to change the type definition of `NumberFormatProps` to: 
```
export interface NumberFormatProps
    extends InputAttributes {
    ...
    defaultValue?: React.HTMLAttributes<any>['defaultValue'];
    ....
}
```

Then this will always be in sync with `@types/react`.

BTW: Seems like there was a similar issue two years ago: https://github.com/s-yadav/react-number-format/issues/244

PR #464, opened 2020-11-23
Added string[] and readonly string[] for NumberFormatProps for defaultValue

Magofoco:
This PR try to fix a type error of `NumberFormatProps` when using the MaterialUi Textfield `inputComponent` in `InputProps`

Failing example:
https://codesandbox.io/s/wonderful-chaplygin-f3otf
If you go to the component `MyInput` you can see that the component `NumberFormatCustom` takes `NumberFormatProps` as props.
The Material UI Textfied, in its  `InputProps`, can take a `inputComponent` which can be the `NumberFormatCustom` component.
This leads to a type error since the props Textfied's `defaultValue` types are: `any` (https://material-ui.com/api/text-field/)
while the NumberFormatProps's defaultValue is defined as: `string | number | undefined`.

I tried to extend the types of `defaultValue` by including `string[]` and `readonly string[]` which seem to solve the above mentioned error

### DIFF
--- a/typings/number_format.d.ts
+++ b/typings/number_format.d.ts
@@ -39,7 +39,7 @@ declare module "react-number-format" {
     removeFormatting?: (formattedValue: string) => string;
     mask?: string | string[];
     value?: number | string;
-    defaultValue?: number | string;
+    defaultValue?: number | string | readonly string[];
     isNumericString?: boolean;
     customInput?: React.ComponentType<any>;
     allowNegative?: boolean;
--- a/typings/number_format.d.ts
+++ b/typings/number_format.d.ts
@@ -39,7 +39,7 @@ declare module "react-number-format" {
     removeFormatting?: (formattedValue: string) => string;
     mask?: string | string[];
     value?: number | string;
-    defaultValue?: number | string | readonly string[];
+    defaultValue?: number | string | string[] | readonly string[];
     isNumericString?: boolean;
     customInput?: React.ComponentType<any>;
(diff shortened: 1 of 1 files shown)


typings/number_format.d.ts
```diff
@@ -39,7 +39,7 @@ declare module "react-number-format" {
     removeFormatting?: (formattedValue: string) => string;
     mask?: string | string[];
     value?: number | string;
-    defaultValue?: number | string;
+    defaultValue?: number | string | string[] | readonly string[];
```
s-yadav, line 42: But isn't this wrong.?  defaultValue will never be array of string.

Magofoco: @s-yadav 
I was thinking the same, but apparently it can literally take `any`.
See here: https://codesandbox.io/s/awesome-benz-7vj06 with different `defaultValue` types

For example, an array of string such as:
`const myArray = ["first", "second", "third"]`
if assigned to `defaultValue` will produce:
`first, second, third`

s-yadav: It can because of typecasting by JS. But that will still be wrong. Also, the type of value and default value should match.
Wrong typing can lead to confusion on dev end.

Is there any other way to force type in material ui.?

I guess the Material-UI has mentioned it as any because the Material UI supports the text field to be rendered as select.

s-yadav:
Related issue.
https://github.com/s-yadav/react-number-format/issues/446

PR #479, opened 2021-01-04
chore: Add CSB Link and browsers to PR Template

nikhil-varma:
## Please check which browsers were used for testing

- [ ] Chrome
- [ ] Safari (OSX)
- [ ] Safari (iOS)
- [ ] Firefox


nikhil-varma:
@s-yadav Added CSB link and a couple more browsers to the PR template. Should we create an issue template as well?

s-yadav:
Yes. The issue template also makes sense.

Issue #495, opened 2021-03-06
18 or more Decimal Scale


Dellybro:
## Describe the issue and the actual behavior

When using a decimal scale of 18+ the decimals i input change. For example, if i am to put in the value 1.12, it fire an onValueChanged with 1.12 and then immediately fires another on Change with 1.11. 

## Describe the expected behavior

I would expect using 18 decimal scale that if i put in 1.12 the value would stay as 1.12 and not update to 1.11

## Provide a CodeSandbox link illustrating the issue

https://codesandbox.io/s/material-demo-forked-6jcgb

## Provide steps to reproduce this issue

- Use Material UI and NumberFormatField as the InputComponent
- Use isNumericString={true}

## Please check the browsers where the issue is seen
- [x ] Chrome
- [ x] Chrome (Android)
- [ x] Safari (OSX)
- [ x] Safari (iOS)
- [ x] Firefox
- [ x] Firefox (Android)

EDIT:

This only happens when isNumericString is set to true

s-yadav:
This is some strange behavior. only happens for `.12`
@ParthS007, @nikhil-varma anyone wants to look into this.?

nikhil-varma:
@s-yadav Sure! I can take a look at this. Shall I pick this up?

nikhil-varma:
@s-yadav I have noticed it is happening for `.12`, `.15` and `.18`....

nikhil-varma:
@Dellybro  @s-yadav This looks like a floating point issue with JS's `toFixed` :

In the `roundToPrecision` function, we are getting the `roundedDecimalParts`. This mis-behaves for a few cases as the video shows below:

<img src="https://im2.ezgif.com/tmp/ezgif-2-f8e1c9bad441.gif"  width="100%" />

The origination is from this line: `Math.ceil(parseFloat("0.".concat(afterDecimal || '0'))).toFixed(scale)` meaning that the `toFixed` method call has issues. This is basically a floating point issue with JS 😝 

Now for values like `.12`, `.15`, `.18` the returned outputs are `.001` less than the actual number. For eg: if the `afterDecimal` value is `12` -in our case, then the assigned value to `roundedDecimalParts[1]` would be `"119999999999999996"` and this would then be passed to `limitToScale` which will cap it to the scale provided. 

I am working on a fix for this and will keep you updated!

Dellybro:
Thank you so much! I realized also if i take away "isNumericString" this works as expected so it has something to do with that boolean as well.

[...]

PR #501, opened 2021-03-13
Fix decimal scale issue

nikhil-varma:
## Describe the issue/change
This PR fixes decimal scale issue. Ref issue linked in this PR

## Add CodeSandbox link to illustrate the issue (If applicable)
https://codesandbox.io/s/material-demo-forked-6jcgb?file=/demo.js:0-2546

## Describe specs for failing cases if this is an issue (If applicable)
When a value like `123.15` is passed, the final value returned and assigned to the input is `123.14`. This is happening because of the `toFixed` method causing a floating point issue. 

## Describe the changes proposed/implemented in this PR
This fix ensure that the scale is capped only when the decimal value overshoots the scale value

## Link Github issue if this PR solved an existing issue
Fix for : #495 

## Example usage (If applicable)
N/A

## Screenshot (If applicable)
N/A

## Please check which browsers were used for testing
- [x] Chrome
- [ ] Chrome (Android)
- [ ] Safari (OSX)
- [x] Safari (iOS)
- [ ] Firefox
- [ ] Firefox (Android)


src/utils.js
```diff
@@ -89,7 +89,8 @@ export function roundToPrecision(numStr: string, scale: number, fixedDecimalScal
 
   const shoudHaveDecimalSeparator = numStr.indexOf('.') !== -1 && scale;
   const {beforeDecimal, afterDecimal, hasNagation} = splitDecimal(numStr);
-  const roundedDecimalParts = parseFloat(`0.${afterDecimal || '0'}`).toFixed(scale).split('.');
+  const floatValue = afterDecimal.length <= scale ? parseFloat(`0.${afterDecimal || '0'}`).toString() : parseFloat(`0.${afterDecimal || '0'}`).toFixed((scale));
+  const roundedDecimalParts = floatValue.split('.');
```
s-yadav, line 93: Too many things happening at ternary. This could be made more readable by breaking it.

```suggestion
  const floatValue = parseFloat(`0.${afterDecimal || '0'}`);
  const floatValueStr = afterDecimal.length <= scale ? floatValue.toString() : floatValue.toFixed(scale);
  const roundedDecimalParts = floatValueStr.split('.');
```

test/library/input.spec.js
```diff
@@ -425,5 +425,40 @@ describe('NumberFormat as input', () => {
         shallow(<NumberFormat format="#### #### ####" mask={['D', 'D', 'M', '1', '2', 'Y', 'Y', 'Y']}/>)
       }).toThrow()
     })
+
+    it('should show the right decimal values based on the decimal scale provided', () =>{
```
s-yadav, line 429: ** Question **
Did this fail without the above change? I tested the behaviour (described it test) in codesandbox with current version and it did not fail.

Try removing your changes in source file and see if this fails.

s-yadav (approved):

PR #529, opened 2021-05-02
fix: add type number for html type atribute

alexanderomnix:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number

My end proposal is to use just string as the type definition for, but I added this pull just adding number as an minimal change.

Typescript releatable: 
When I use the prop type number on the NumberFormat component I got a type error allowing just text | tel | undefined, this is a minimal change to accept the number on type prop as well.
 You guys can just use as string to accept any string pass thought if this lib is just using this props to inject into the dom element.

No applicable
- [ ] Chrome
- [x] Chrome (Android)
- [x] Safari (OSX)
- [ ] Safari (iOS)
- [ ] Firefox
- [ ] Firefox (Android)

### DIFF
--- a/typings/number_format.d.ts
+++ b/typings/number_format.d.ts
@@ -54,7 +54,7 @@ declare module "react-number-format" {
      * onFocus: Function;
      * onBlur: Function;
      */
-    type?: 'text' | 'tel'  | 'password';
+    type?: 'text' | 'tel'  | 'password' | 'number';
     isAllowed?: (values: NumberFormatValues) => boolean;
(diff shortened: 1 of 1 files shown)


s-yadav:
The problem with passing type as number is it will not support any formatting characters in between. And if you don't need formatting character, using normal input will be a better choice.

nikhil-varma:
@alexanderomnix Please let us know if you need any further assistance on any issue. Considering this, I am closing this PR as @s-yadav suggested above that this approach may have potential pitfalls. Feel free to comment here or create an issue if you need any further help

PR #535, opened 2021-05-11
feat: added support for persian numeral input

electather:
added support for Persian numeral which automatically formats it to standard English numeral 

try inputting ۱۲۳۴۵۶۷۸۹۰ in the input

none

added an optional prop `convertPersianDigits`  which triggers a customized regex and a formatter 

none

try inputting ۱۲۳۴۵۶۷۸۹۰ in the input

- [ *] Chrome
- [* ] Chrome (Android)
- [ ] Safari (OSX)
- [ ] Safari (iOS)
- [ *] Firefox
- [ ] Firefox (Android)


s-yadav:
Instead of adding checks specific for Persian input, I think we should allow user to provide the numeric regex as props default  to [0-9].

s-yadav:
Also, @nikhil-varma Can we run prettier on all files once, so the PR doesn't include the prettier changes.

nikhil-varma:
@s-yadav sure! Sounds good

electather:
@s-yadav so, providing a prop like inputRegex that defaults to /d and a function that transforms that input to English numerals.
or we can do it in an array like:
```ts
{
  regex: string,
  transformer : () => number;
}
```

s-yadav:
@electather Sorry for replying late. But I didn't get the use case of providing transformer. This can remain outside of NumberFormat right?

electather:
One place that it could've helped is the onValueChange where we have the parsed values. If user decides to override the regex then the input cannot be converted to int and user can only receive the formatted string value. 
But forcing the user to provide transformer on regex override will prevent this issue

s-yadav:
Make sense, not just on onValueChange. Internally we use numbers to round to define precision. So conversion is indeed needed. 
The other approach to do this can be.
```js
customNumerals: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9] // where people can provide numeric mapping based on the numeral they want to support.
```
With this we can create a regex. `new RegExp(customNumerals.join('|'))`
And can also apply the transformation in the lib only.

The only problem is what if the customNumerals do not follow character to character map, like Roman Numerals. People might expect this numeral as well to work which would be very tricky to implement.

electather:
@s-yadav seems like a great idea. i'm gonna close this pull request and start working on a new one. ( since it has little in common with what is currently purposed)

PR #536, opened 2021-05-12
types: export index signature (catch-all) prop types separately

Bogidon:
#### Describe the changes proposed in this PR
This PR introduces a minor, backwards-compatible change to the TypeScript typings. It extracts the known props of `NumberFormatProps`: everything but the index signature (`[key: string]: any`), into a new interface `NumberFormatPropsBase`. The existing `NumberFormatProps` remains the same.

#### Describe the issue
What does this change accomplish? Authors of consuming packages/applications wanting to maintain strict TypeScript checking may wish to not allow props other than those of known to be explicitly supported to be passed to their react components.

My organization takes such a stance, so we maintain a list of allowed props by defining a separate interface:
```ts
type FormattedInputProps = Pick<NumberFormatProps, 'allowNegative'  | 'format' | '...'>
```
However this interface is lengthy and a maintenance burden. If we could import NumberFormatPropsBase instead, we wouldn't need to do this whitelisting ourselves.

#### Alternatives considered
If the only purpose of the catch-all index signature is to pass unknown props to a customInput component if one is provided, generics _could_ be used to properly type everything. This took me more time than I'm willing to admit to figure out how. The method is using the latest TypeScript features from version 4.1+, so it's probably a non-starter. In any case here's a Code Sandbox I made as a proof of concept of this approach: https://codesandbox.io/s/typescript-playground-export-forked-31cnw?file=/index.tsx

##### Changelogg
 - (feature) add NumberFormatPropsBase to export prop types without index signature


s-yadav:
This looks good. For customInput, currently we are passing any. Can we define generics default to any.?
Also. There is some conflict due to prettier change. Can you check that as well.

Bogidon:
@s-yadav updated to fix the conflict. I did not end up using generics in the PR for the reasons listed.

But in exploration that I did, which would indeed be a breaking change (an additional reason I am not proposing it), I'm not sure you would want to default to `any`, because the purpose was to derive the props from the type of CustomInput – if there is no customInput, you probably do not want extra allowed props. Consumers can pass a `React.FC<any>` if they want `any` props.

I suggest bringing this non-breaking change in first and then exploring the use of generics separately.

s-yadav (approved):

s-yadav:
Sure, approved.

So I was asking something like this, until we get your POC working. 
```
interface NumberFormatProps<T = any> {
    ...
    constomInput: React.ComponentType<T>
}
```

Currently it is
```
interface NumberFormatProps {
    ...
    constomInput: React.ComponentType<any>
}
```

But not sure, if this will provide any value. as typescript will not know what props belongs to customInput.

Bogidon:
Ah, yeah. That would be possible, but correct I don't think it would really provide value without using `T` to derive the passed props.

s-yadav:
@Bogidon. Merged. This will go on next release. Will try to do a minor release soon.

PR #546, opened 2021-05-30
Align Docs Is Allowed Example With Type Definition

alexanderdavide:

The [docs](https://github.com/s-yadav/react-number-format/blob/master/README.md) show a wrong example of the `isAllowed` callback. It returns the `inputObj` if it passes the check or `undefined`. This actually works in JavaScript because of type coercion (IMO still a bad practice). However, in TypeScript the type definition enforces a `boolean` to be returned.

```javascript
const MAX_VAL = 1400;
const withValueLimit = (inputObj) => {
  const { value } = inputObj;
  if (value < MAX_VAL) return inputObj;
};
<NumberFormat value={12} isAllowed={withValueLimit} />;
```

The [type definition](https://github.com/s-yadav/react-number-format/blob/master/typings/number_format.d.ts) requires a `boolean`:
```typescript
isAllowed?: (values: NumberFormatValues) => boolean;
```

Aligned the `isAllowed` example in the docs with the method's type definition.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,6 +211,8 @@ function cardExpiry(val) {
   <NumberFormat value={12} isAllowed={withValueLimit} />;
 ```
 
+We are using `floatValue` here because it provides the exact value (see [values object](#values-object)). `value` is rounded to your configured `decimalScale`; hence it may omit a part of the precision of the actual value. Given an entered value of 1400.4 and the default `decimalScale`, `value` would be transferred as 1400 in the values object and pass the above `withValueLimit` check. In this case, it is a false positive that can be prevented using `floatValue`. However, you might want to use `value` to automatically get the rounded number according to your `decimalScale`.
+
 Visit this link for Demo: [Field with value limit](https://codesandbox.io/s/react-number-format-isallowed-8gu0v)
 
 ![Screencast example](https://media.giphy.com/media/7agdv1sb9JYJEYlN3s/giphy.gif)
--- a/README.md
+++ b/README.md
@@ -209,7 +209,8 @@ function cardExpiry(val) {
   const MAX_VAL = 1400;
   const withValueLimit = (inputObj) => {
     const { value } = inputObj;
-    if (value < MAX_VAL) return inputObj;
+    if (value >= MAX_VAL) return false;
+    return true;
   };
   <NumberFormat value={12} isAllowed={withValueLimit} />;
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -209,7 +209,8 @@ function cardExpiry(val) {
   const MAX_VAL = 1400;
   const withValueLimit = (inputObj) => {
     const { value } = inputObj;
-    if (value < MAX_VAL) return inputObj;
+    if (value >= MAX_VAL) return false;
```
nikhil-varma, line 212: Nice catch! @alexanderdavide We can simplify this into something like this
```
const withValueLimit = inputObj => value <= MAX_VAL;
```
This should return either `true` or `false` and we may not need explicit return statements. Also, I guess it should be `value <= MAX_VAL` since `2 <= 2` is also a valid case 🤔

nikhil-varma (changes requested):

README.md
alexanderdavide: Thank you for the recommendations. Happy to use shorthands, but initially opted for a more explained version. You're totally right with the less than sign. I mixed things up.

I have updated the PR. Furthermore, I think it's better to rely on `floatValue` because you have the exact value to do comparisons on. `value` is rounded; hence it omits the precision part of the value. Think about an actual value of 1400.4: It would be transferred as 1400 in `value` and pass the check which is a false positive. This can be prevented using `floatValue`.

nikhil-varma: @alexandrapersea Sounds great! 

One suggestion here.. let's add both the examples.. one with `value` and one with `floatValue`. That way it will be very evident about what can be done and it's a good to have since it's always a good idea for documentation to be extensive.  Thanks again!

alexanderdavide: I have added an explanation outlining the use cases. In my opinion, the text makes it very clear. I decided against adding code for both `value` and `floatValue` to not make it too verbose.

nikhil-varma: @alexandrapersea Sounds good. That works too. Thanks!

alexanderdavide: Great. Would you then undo the change request so that we can complete this PR?

nikhil-varma: @alexanderdavide Done!

nikhil-varma (approved):

nikhil-varma:
@alexanderdavide Merging this PR

Issue #533, opened 2021-05-07
All zero string decimals (e.g., '100.00') strips the '.00' and displays as '100.'


codypearce:
#### Describe the issue and the actual behavior
Values in the app I am working on are passed like this '100.00', with a `decimalScale` of 2, and `isNumericString`, like this.

```jsx
<NumberFormat
        name={"AmountCents"}
        value={"100.00"}
        thousandSeparator
        prefix="$"
        decimalScale={2}
        isNumericString
      />
```

The result in the app 
<img width="207" alt="Screen Shot 2021-05-07 at 4 39 22 AM" src="https://user-images.githubusercontent.com/12564956/117444350-36d04380-aeee-11eb-8d26-82b71512b073.png">

#### Describe the expected behavior
Since it was passed in as `100.00` it should display as `100.00`. 

The changes in #501 appear to be the issue. 
```js
var floatValue = parseFloat("0.".concat(afterDecimal || '0'));
var floatValueStr = afterDecimal.length <= scale ? floatValue.toString() : floatValue.toFixed(scale);
```
`floatValue` is `0`
`afterDecimal.length` is 2 and `scale` is 2
so `floatValueStr` returns `floatValue.toString()` which returns `0`
However, `floatValueStr` should probably return `0.00`

#### Provide a CodeSandbox link illustrating the issue
https://codesandbox.io/s/vigilant-shamir-dk4u9?file=/src/App.js

#### Provide steps to reproduce this issue
See codesandbox

#### Please check the browsers where the issue is seen
- [x] Chrome
- [ ] Chrome (Android)
- [x] Safari (OSX)
- [ ] Safari (iOS)
- [ ] Firefox
- [ ] Firefox (Android)

nikhil-varma:
Hey @codypearce Thanks for reporting this issue. We will investigate and shoot out a fix soon. Keep an eye out for updates 🙂

nikhil-varma:
@codypearce Just a quick work around until the issue is being investigated. You can use the `fixedDecimalScale` prop to match the leading zeros.  Here's a sample Codesandbox link: https://codesandbox.io/s/zealous-oskar-wjnqh

From the document for `fixedDecimalScale` 
> If true it add 0s to match given decimalScale

I will get back to you with an update on the issue resolution. Thank you for your patience.

nikhil-varma:
@codypearce Update on the issue

[...]

PR #550, opened 2021-06-08
Fix issue #533

nikhil-varma:
This fixes the trailing zeros issue when an input like `###.00` is provided

N/A

- Should show the right number of zeros in all cases based on the `decimalScale` provided

- Adds a function to identify if the string is all zeros
- Uses that conditionally to pass the value to `limitToScale`

Issue: #533 

```
<NumberFormat
        name={"AmountCents"}
        value={"100.00"}
        thousandSeparator
        prefix="$"
        decimalScale={2}
        isNumericString
      />
```

Output: 100.00

- [x] Chrome
- [ ] Chrome (Android)
- [ ] Safari (OSX)
- [ ] Safari (iOS)
- [ ] Firefox
- [ ] Firefox (Android)

### DIFF
--- a/test/library/input.spec.js
+++ b/test/library/input.spec.js
@@ -554,8 +554,10 @@ describe('NumberFormat as input', () => {
       expect(wrapper.find('input').instance().value).toEqual('$100.0');
       wrapper.setState({ value: '123.00' });
       expect(wrapper.find('input').instance().value).toEqual('$123.00');
-      wrapper.setState({ value: '100.000' });
-      expect(wrapper.find('input').instance().value).toEqual('$100.00');
+      wrapper.setState({ value: '132.000' });
+      expect(wrapper.find('input').instance().value).toEqual('$132.00');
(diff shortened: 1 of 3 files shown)


test/library/input.spec.js
```diff
@@ -524,5 +526,38 @@ describe('NumberFormat as input', () => {
       wrapper.setState({ value: '123.1234' });
       expect(wrapper.find('input').instance().value).toEqual('$123.1234');
     });
+
+    it('should show the right number of zeros in all cases', () => {
```
s-yadav, line 530: Can you add the issue number in the testcase.

s-yadav (approved):

test/library/input.spec.js
s-yadav: May be as comment.

nikhil-varma: @s-yadav Done!

Issue #349, opened 2019-10-02
allowDecimalSeparators does't always work correctly when paste a number into the input


tenkij:
decimalSeparator === '.'
1) paste '11.11', result '11.11';
2) paste '11,11', result '1 111'

decimalSeparator === ','
1) paste '11.11', result '1 111';
2) paste '11,11', result '11,11'

https://codesandbox.io/s/react-number-format-demo-f7245

dfyz011:
i have same issue

obrejla:
Hi. Same issue here...:/

ArtemShirshov:
Hi, @s-yadav 
This problem is still relevant today

farmer911:
@tenkij thank for your code-pen i had fixed the issue. Love you 3000

LeComptoirDesPharmacies:
Hi @s-yadav,
Same problem here ;)

jafidiesel:
I'm not very sure but i think maybe it's forcing formatting the value since it's a number greater than a thousand. By default if thousandSeparator prop it's not present it shouldn't force formatting, but it's a posibility.

Toumash:
Still relevant today

nikhil-varma:
@Toumash That would be great! Let me know how I can help you with this! Thanks!

Toumash:
@nikhil-varma You can review the [attached PR](https://github.com/s-yadav/react-number-format/pull/556) :)

Toumash:
For those who are waiting for the PR to be merged: you can use customInput

```js
const DemoField =  useMemo(() => NumberFormatCommaPasteHackTextField(setValue), []);

// HACK: replace with normal TextField after this PR is merged https://github.com/s-yadav/react-number-format/pull/556
const NumberFormatCommaPasteHackTextField = (setValue) =>
  (props) => {
    return <TextField
      {...props}
      onPaste={(e) => {
        let pastedText = e.clipboardData.getData('text');
        if (pastedText.indexOf(',') !== -1) {
          e.preventDefault();
          setValue(pastedText.replace(',', '.'));
        }
      }}
    />
  }

return  <NumberFormat
        customInput={DemoField}
        allowedDecimalSeparators={[",", "."]}
        decimalSeparator={'.'}
        thousandSeparator=" "
      />
```

dainius-tamosiunas-tg:
@s-yadav any news on this?

PR #556, opened 2021-06-23
add support for paste event given allowedDecimalSeparators

Toumash:
Described here https://github.com/s-yadav/react-number-format/issues/349 by @tenkij
[Codesandbox by @tenkij](https://codesandbox.io/s/react-number-format-demo-f7245)
allowedDecimalSeparators={[",", "."]}

decimalSeparator='.'

1. paste '11.11', result '11.11';
2. paste '11,11', result '1 111'

decimalSeparator=','

1. paste '11.11', result '1 111';
2.  paste '11,11', result '11,11'

Now the paste works with both cases

https://user-images.githubusercontent.com/9840635/123081289-1a488600-d41e-11eb-8a97-002850e185bd.mp4

- [x] Chrome
- [ ] Chrome (Android)
- [ ] Safari (OSX)
- [ ] Safari (iOS)
- [ ] Firefox
- [ ] Firefox (Android)

### DIFF
--- a/example/src/index.js
+++ b/example/src/index.js
@@ -114,6 +114,16 @@ class App extends React.Component {
         </div>
 
         <div className="example">
+          <h3>Custom allowed decimal separators with decimal precision</h3>
+          <NumberFormat
+            thousandSeparator={' '}
+            allowedDecimalSeparators={['.', ',']}
+            decimalSeparator='.'
+            decimalScale={2}
+          />
+        </div>
+
+        <div className="example">
           <h3>Format with pattern : Format credit card in an input</h3>
           <NumberFormat format="#### #### #### ####" mask="_" />
         </div>
--- a/src/number_format.js
+++ b/src/number_format.js
@@ -671,6 +671,16 @@ class NumberFormat extends React.Component {
       );
     }
 
+    /** Check for "paste event" with, replace any allowed decimal separator with desired decimal separator */
+    if (!format && decimalScale !== 0 && allowedDecimalSeparators.some(separator => value.indexOf(separator) !== -1)) {
+      const separator = decimalScale === 0 ? '' : decimalSeparator
+      let result = value;
+      allowedDecimalSeparators.forEach(v => {
+        result = result.replace(v, separator);
+      })
+      value = result;
+    }
+
     const leftBound = !!format ? 0 : prefix.length;
     const rightBound = lastValue.length - (!!format ? 0 : suffix.length);
 
--- a/test/library/input_numeric_format.spec.js
+++ b/test/library/input_numeric_format.spec.js
@@ -603,6 +603,22 @@ describe('Test NumberFormat as input with numeric format options', () => {
     expect(wrapper.state().value).toEqual('23,4456 Sq. ft');
   });
 
+
+  it('should allow pasting 123,45 with decimal separator set to . and allowedDecimalSeparators=["," , "."]', () => {
+    const wrapper = shallow(
+      <NumberFormat
+        displayType="input"
+        thousandSeparator={' '}
+        prefix=""
+        allowedDecimalSeparators={['.', ',']}
+        decimalSeparator='.'
+        decimalScale={2}
+      />,
+    );
+    simulateKeyInput(wrapper.find('input'), '123,45');
+    expect(wrapper.state().value).toEqual('123.45');
+  });
+
   it('should not break if suffix/prefix has negation sign. Issue #245', () => {
     const wrapper = shallow(<NumberFormat suffix="-" />);
 


src/number_format.js
```diff
@@ -671,6 +671,16 @@ class NumberFormat extends React.Component {
       );
     }
 
+    /** Check for "paste event" with, replace any allowed decimal separator with desired decimal separator */
+    if (!format && decimalScale !== 0 && allowedDecimalSeparators.some(separator => value.indexOf(separator) !== -1)) {
+      const separator = decimalScale === 0 ? '' : decimalSeparator
```
nikhil-varma, line 676: In the previous line we have a constraint for `decimalScale` to not be 0 so `decimal === 0` seems like it will never be true.. Can we fix this?

example/src/index.js
```diff
@@ -113,6 +113,16 @@ class App extends React.Component {
...
         </div>
 
+        <div className="example">
+          <h3>Custom allowed decimal separators with decimal precision</h3>
+          <NumberFormat
+            thousandSeparator={' '}
+            allowedDecimalSeparators={['.', ',']}
+            decimalSeparator='.'
+            decimalScale={2}
+          />
```
nikhil-varma, line 123: Let's keep this example simple by just adding it like this
```
 <NumberFormat
            value={11.11}
            allowedDecimalSeparators={['.', ',']}
            decimalSeparator='.'
            decimalScale={2}
          />
```

Because pasting may not be immediately evident so it's a good idea to have some value and let Devs'/Users' play with it

test/library/input_numeric_format.spec.js
```diff
@@ -603,6 +603,22 @@ describe('Test NumberFormat as input with numeric format options', () => {
     expect(wrapper.state().value).toEqual('23,4456 Sq. ft');
   });
 
+
+  it('should allow pasting 123,45 with decimal separator set to . and allowedDecimalSeparators=["," , "."]', () => {
```
nikhil-varma, line 607: Let's rename this to 
```suggestion
  it('should work with the configured decimal separators based on the allowedDecimalSeparators prop', () => {
```

nikhil-varma (commented):
Can you also please add a related documentation in README.md file for this prop? 

Also, although things look good, I would still prefer another review done by @s-yadav to ensure all perspectives are considered and we don't miss anything. I appreciate your patients on this on this! Thanks!

Toumash:
> Can you also please add a related documentation in README.md file for this prop?
> 
> Also, although things look good, I would still prefer another review done by @s-yadav to ensure all perspectives are considered and we don't miss anything. I appreciate your patients on this on this! Thanks!

Done. Waiting for the second review :)

src/number_format.js
```diff
@@ -671,6 +671,15 @@ class NumberFormat extends React.Component {
...
     }
 
+    /** Check for "paste event" with, replace any allowed decimal separator with desired decimal separator. Issue #349  */
+    if (!format && decimalScale !== 0 && allowedDecimalSeparators.some(separator => value.indexOf(separator) !== -1)) {
+      let result = value;
+      allowedDecimalSeparators.forEach(v => {
+        result = result.replace(v, decimalSeparator);
+      })
+      value = result;
+    }
```
s-yadav, line 681: - The logic can be simplifed using regex, instead of array.
```
const decimalSeparatorRegex = new RegExp(allowedDecimalSeparators.join('|'));
```
- If we add this the previous `if` condition is no longer needed.
- The replace method should happen only on characters between prefix and suffix. What if prefix or suffix has a decimal separator character. like suffix `.sqft`. Can we add a spec to verify that? <- Add decimal char on the prefix, as replace replaces first char.

s-yadav (changes requested):

src/number_format.js
s-yadav: Also, while looking into code, I feel this transformation is non needed to be done on `correctInputValue`. Instead we should do it in formatInput method, after this.
https://github.com/s-yadav/react-number-format/blob/master/src/number_format.js#L621

That way we don't have to worry about prefix and suffix. Also, we can remove transformation for allowedDecimalSeparator in line 663.
https://github.com/s-yadav/react-number-format/blob/master/src/number_format.js#L663

s-yadav: correctInputValue, is mostly a supporting function for keyDown event. This was introduced due to android keyboard bug, where it doesn't give the correct character code on the keyDown event. Not sure if it still exist.

s-yadav: On further discussion looks like the paste event is tricky, what if the pasted text have a thousand separator on it. For example, `1,111.11` This logic will break.

Also, how do you identify if `,` in pasted character is supposed to be thousand separator or decimal separator. For example what we should treat `,` in this. `1,111`.?

Toumash: > On further discussion looks like the paste event is tricky, what if the pasted text has a thousand separator on it. For example, `1,111.11` This logic will break.
> 
> Also, how do you identify if `,` in pasted character is supposed to be thousand separator or decimal separator. For example what we should treat `,` in this. `1,111`.?

As far as i know you cannot have allowedDecimalSeparator same as thousand separator. I believe decimal separator difference is more common than thousand separator differences.
When you look at the current library props it looks like you can enter different decimal place separator, but only one thousand separator - and that should stay that way

nikhil-varma: Hey, just one small clarification on this _(assuming this is what you meant)_

> As far as i know you cannot have allowedDecimalSeparator same as thousand separator. I believe decimal separator difference is more common than thousand separator differences.

At the moment, we cannot have decimal separator same as the thousand separator. Here is a small demo:           https://codesandbox.io/s/react-number-format-demo-forked-gpm7s
It should throw this error: `Decimal separator can't be same as thousand separator.`

This, -on the other hand, is possible

```
 <NumberFormat
        allowedDecimalSeparators={[",", "."]}
        decimalSeparator="."
        isNumericString
        thousandSeparator=","
      />
```

For some more detail on `allowedDecimalSeparators` you can refer this: #345

Toumash: Do we really want to support this configuration with thousandSeparator=',' and allowedDecimalSeparators={[",", "."]} ?

s-yadav: I think there is valid usecase for that configuration.
https://github.com/s-yadav/react-number-format/issues/324#issuecomment-534439760

s-yadav: Other people also to support similar usecase. If you go through through linked issues in this PR. #345 

The allowedDecimalSeparator was introduced to aid on keyboards where locale decides what char to show. Paste is tricky as mentioned above.

One thing we can do is to run the logic only when thousandSeparator doesn't conflict with allowdDecimalSeparator.

This will not solve the problem completely but partial improvement is better than no improvement.

Thoughts @Toumash, @nikhil-varma

nikhil-varma: I think considering what `allowDecimalSeparator` is doing, a validation agains `thousandSeparator` might be a good idea. On that note, there are one of two things we can choose to do
- Add a base check to ensure that `thousandSeparator` is not included in `allowDecimalSeparator` _(Rationale is basically that a string cannot have the same radix character and thousand separator `_ , Ideally _(in the context of our library)_ the radix character is dependent on either
  - The value of `decimalSeparator` 
  - Default `.`
- Just consider this check while pasting 

I am guessing adding it as a base condition is a much cleaner idea because the library will be able to tell the users' that it is not a valid config and will provide a clear segregation unless we have a use-case where it might fail. Because in the end, a value in the `thousandSeparator` is an allowed character but not as a radix character. 

What do you think @s-yadav @Toumash

s-yadav: Validation might not be possible as there is a valid usecase for the conflict.

https://github.com/s-yadav/react-number-format/issues/324#issuecomment-534439760

nikhil-varma: Oh yeah you are right! But then if we just add it in this particular context, this specific use-case might fail? 🤔

Toumash: Okay, maybe we should just add this workaround into the README.md then?

```
const DemoField =  useMemo(() => NumberFormatCommaPasteHackTextField(setValue), []);

const NumberFormatCommaPasteHackTextField = (setValue) =>
  (props) => {
    return <TextField
      {...props}
      onPaste={(e) => {
        let pastedText = e.clipboardData.getData('text');
        if (pastedText.indexOf(',') !== -1) {
          e.preventDefault();
          setValue(pastedText.replace(',', '.'));
        }
      }}
    />
  }

return  <NumberFormat
        customInput={DemoField}
        allowedDecimalSeparators={[",", "."]}
        decimalSeparator={'.'}
        thousandSeparator=" "
      />
```

nikhil-varma: Adding it to the documentation may be an issue because it would become a suggested approach and the actual issue won't be resolved. 
Let's do this
- Change documentation for `allowedDecimalSeparators` saying that paste may not work as expected if there is a conflict between `thousandSeparator` and `allowedDecimalSeparators`. Let's also link this PR there in the documentation to provide the rationale behind it.
- We can add a conflict check in your logic between `thousandSeparator` and `allowedDecimalSeparators` as @s-yadav pointed out. 

If there is no conflict then the paste will work as expected but if there is it may not as suggested above.

Toumash: Okay, gonna provide a new version in the next days

sowtame (approved):

Misiu:
@s-yadav can we ask for your review on this? This would be a very helpful feature. Thank you

sowtame:
any updates? )

rhmnaulia:
any updates?

Toumash:
Nope, feel free to take it over

Toumash:
> Are there any plans to merge this?

Nope, feel free to take it over

Issue #422, opened 2020-07-01
Typescript error when pass "size" prop with custom MUI Textfield


Skladnayazebra:
Hello!

I`m using NumberFormat + Material UI + TypeScript

When i give `customInput={TextField}` and `size={"small"}`, I got error from TypeScript:

```
No overload matches this call.
Overload 1 of 2, '(props: Readonly<NumberFormatProps>): NumberFormat', gave the following error.
    Type 'string' is not assignable to type 'number'.
  Overload 2 of 2, '(props: NumberFormatProps, context?: any): NumberFormat', gave the following error.
    Type 'string' is not assignable to type 'number'.  TS2769
```
Here`s my component:

```
<NumberFormat
    format={'############'}
    mask={'_'}
    onValueChange={(values: NumberFormatValues) => {
        setValue('identification.inn', values.value)
    }}
    customInput={TextField}
    size={"small"}
    label={'ИНН'}
    variant="outlined"
    error={Boolean(errors.identification?.inn)}
    helperText={'12 цифр'}
    classes={{root: styles.textField}}
/>
```

If I remove `size` prop, everything works fine and `TextField` recieve all other props (`label`, `helperText`, etc.)

S3ak:
I have the same issue. How did you resolve?

S3ak:
Hard workaround is to ignore the props type checking
```
// @ts-ignore
size="small
```

j0n4t:
Another workaround is making a react function with the needed props:
`
const SmallTextField = (props: Object) =>
  <TextField size='small' {...props} />
`

And then, passing it to customInput:
`<NumberFormat  customInput={SmallTextField}   />`

silvergraphs:
Please fix it

> Another workaround is to make a react function with the needed props:
> `const SmallTextField = (props: Object) => <TextField size='small' {...props} />`
> 
> And then, passing it to customInput:
> `<NumberFormat customInput={SmallTextField} />`

Not worked for me

dylancrockett:
Since this still has not been fixed I figured I would share the solution I came up with in a project I am currently working on. I hope this will be fixed soon but in the meantime I figured I would share what I managed to come up with.

Here it is:
```ts
type NumberTextFieldProps = Omit<NumberFormatPropsBase, "size"> & Omit<TextFieldProps, "value" | "defaultValue" | "ref" | "type">;

[...]

PR #577, opened 2021-09-10
feat: make typings for customInput generic

patsimm:

We are using react-number-format together with material-ui. In our case we want to use the custumInput "TextField" but pass the prop: `{ size: "small" }`.

When we do this we get the error:

```
No overload matches this call.
  Overload 1 of 2, '(props: NumberFormatProps | Readonly<NumberFormatProps>): NumberFormat', gave the following error.
    Type 'string' is not assignable to type 'number | undefined'.
  Overload 2 of 2, '(props: NumberFormatProps, context: any): NumberFormat', gave the following error.
    Type 'string' is not assignable to type 'number | undefined'.ts(2769)
```

The problem here is, that `NumberFormatProps` are typed to always extend `InputAttributes`. Where actually when we use a customInput there is no reason to pass `InputAttributes`. In that case we should be able to pass the props of the customInput.

As far as I understand this is how the actual logic works.

https://codesandbox.io/s/beautiful-tess-eb5r6?file=/src/App.tsx

I updated the typings of NumberFormatProps to be generic. Now instead of always extending from `InputAttributes` the Props-Type of the component passed as customInput are extracted and added to the typing. Typescript is actually smart enought to this automatically. 

I must say that I can not promise this does not introduce breaking changes for people that import and use the current typings. But in my opinion those generic typings are closer to how the component actually works.

#422

See the updated typings/number_format.spec.tsx https://github.com/s-yadav/react-number-format/commit/c9e06279aa1fc9fac5b35930de08d508ada6de74#diff-9d8e8f0103177d8c0688a00a41889511f17c18bfa6150b4d4922aa40fc87825aR7

### DIFF
--- a/typings/number_format.d.ts
+++ b/typings/number_format.d.ts
@@ -23,7 +23,7 @@ declare module 'react-number-format' {
     data: any;
   }
 
-  export interface NumberFormatPropsBase extends InputAttributes {
+  export type NumberFormatPropsBase<T> = {
     thousandSeparator?: boolean | string;
     decimalSeparator?: string;
     thousandsGroupStyle?: 'thousand' | 'lakh' | 'wan';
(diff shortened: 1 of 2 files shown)


s-yadav (approved):

s-yadav:
Sorry for looking late into this. Looks great, thanks for PR.

Issue #587, opened 2021-10-25
Listen to input changes event only, while retaining access to the raw values in the callback


DekelYossef:
Hi,

I'm trying to listen to user changes explicitly, and not to every value change due other event types or prop changes, and to get the `values.value` or `values.floatValue` in a callback for easy parsing and handling.

What would be the right way to do so?

Since I can't access the event object in `onValueChange`, and can't find out what's the source of the `onValueChange` which   gets triggered for multiple reasons, I can't use it. Other option I see is using the input's onChange handler, but I get the formatted value from it which requires de-formatting to actually parse the value.

I re-read the documentation, and searched for similar issues, and all I managed to find was https://github.com/s-yadav/react-number-format/issues/429, and more specifically, this:
> Hello, i have approximately the same problem. *___When the props value change the onValueChange is fired___*!! It creates infinite loop when an input update another input vice versa.

_Originally posted by @alexandre033 in https://github.com/s-yadav/react-number-format/issues/429#issuecomment-669152243_

My case is a bit different since I don't update the input value's state and thus there is no infinite loop, but I do get the `onValueChange` triggered when I don't want it to, and can't filter out the specific events I want to listen to.

I would have expected that either I get the `values` object in the `onChange` callback or get the source / reason / trigger for the `onValueChange`. Is there something I'm missing?

Thanks! 🙏🏼

s-yadav:
Humm, the reason we don't want to do it onChange is that, onChange doesn't get called on all cases. Like value can change due to multiple source, prop change, input change, blur. so onChange might not receive all the updates. Also changing the semantic of onChange shouldn't be done. 

But I think we can add the source information in the second param of onValueChange, like mentioned it
https://github.com/s-yadav/react-number-format/issues/429#issuecomment-731706973

I will add this, as this is most asked requirement.

s-yadav:
@nikhil-varma, can you look into this.?

nikhil-varma:
@s-yadav sure!

nikhil-varma:
@DekelYossef We have extended the `onValueChange` callback to allow sending source details as discussed! Please try it out and let us know if you have any further questions. Closing this issue as of now

DekelYossef:
Will do! thank you guys!

PR #591, opened 2021-11-07
feat: Add source details in onValueChange

nikhil-varma:
This PR adds a source object in the `onValueChange` function which returns the event object and the source of the trigger

API change for `onValueChange` to allow getting the source details and the event object

Issue: #587 

```
<NumberFormat
  value={this.state.profit}
  thousandSeparator={true}
  prefix={'$'}
  onValueChange={(values, sourceInfo) => {
    const { formattedValue, value } = values;
    // Event is a Synthetic Event wrapper which holds target and other information. Source tells whether the reason for this function being triggered was an 'event' or due to a 'prop' change
    const { event, source } = sourceInfo;
  }}

```
#### Screenshot (If applicable)#### Please check which browsers were used for testing
- [x] Chrome
- [ ] Chrome (Android)
- [ ] Safari (OSX)
- [ ] Safari (iOS)
- [x] Firefox
- [ ] Firefox (Android)


nikhil-varma:
@s-yadav API changes are done. Any suggestions on what kind of documentation examples we can add? I think we can give more details. What do you say?

s-yadav (approved):

PR #599, opened 2021-11-30
- Fixed #597 (Handle backspace before prefix, when negative value is …

### LINKED ISSUE #597: deleting negative sign with backspace also deletes the last digit

When trying to delete the minus sign '-' with a backspace, the last digit is also removed
Affects inputs with prefixes.

Only the negative sign should be removed

happens even on the official demo site
https://jinno.io/app/23/isAllowed?source=react-number-format&displayType=input

have an input field with the prefix (e.g $)
type a negative number, e.g. -123
you will see something like this: -$123
move cursor left to the front, e.g. before '123' 
press backspace on the keyboard to delete the negative sign

The result: negative sign is removed but also the last digit is removed (e.g. $12) and the cursor is moved to the end

- [x] Chrome

s-yadav:
…present)

- Integrate testing-library

- [ ] Chrome
- [ ] Chrome (Android)
- [ ] Safari (OSX)
- [ ] Safari (iOS)
- [ ] Firefox
- [ ] Firefox (Android)

### DIFF
--- a/test/library/keypress_and_caret.spec.js
+++ b/test/library/keypress_and_caret.spec.js
@@ -1,6 +1,8 @@
 import React from 'react';
 import NumberFormat from '../../src/number_format';
 import ReactDOM from 'react-dom';
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 
 import {
   simulateKeyInput,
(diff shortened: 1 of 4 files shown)


test/library/keypress_and_caret.spec.js
```diff
@@ -249,19 +251,29 @@ describe('Test keypress and caret position changes', () => {
...
-        prefix: '$',
-        value: '-$1,000',
-        onValueChange: spy,
-      });
-      wrapper.update();
 
-      simulateKeyInput(wrapper.find('input'), 'Backspace', 2, 2, setSelectionRange);
-      expect(wrapper.state().value).toEqual('$1,000');
-      expect(caretPos).toEqual(1);
+      const wrapper = render(
```
nikhil-varma, line 257: Won't this return a view? Maybe we can change the name to `const view`

nikhil-varma (approved):
Everything looks good. Just one minor change request.

Issue #611, opened 2022-02-10
remove trailing zeros but keep at least N decimal positions while using thousandSeperator


GTadeon:
So say that I have NumberFormat element with the following props:

```
isNumericString:false
thousandSeparator:true
decimalsScale: 5
fixedDecimalScale:true
```

and I want to be able to cut trailing zeros in the decimal part such that it always displays at least two zeros in the decimal part (which is something you would see in case you have decimalScale:2 anyway), but cuts the rest, bearing in mind that max allowed decimals is 5. How should I go about it ?

so desired output would be:
1 => 1.00
1234 => 1,234.00
123.1234 => 123.1234
12345678.90 => 123,456,78.90
12345678.90000 => 123,456,78.90

I know for the custom format, but what if I don't want to reinvent the wheel and add thousandSeperator  and fixedDecimalScale props' logic by myself (because yeah, I'm losing those two if I go for custom formatting) 

how should I go about this ?

PR #614, opened 2022-02-11
cutTrailingZeros prop added to the component

IGrobenskiTT:
cutTrailingZeros (number) => if passed, then it will look for, and cut, **ALL trailing zeros in the decimal part of the number **, stopping at _cutTrailingZeros_ decimal point. 

So, _cutTrailingZeros_ is also used to determine the number of decimal places to leave intact when cutting trailing zeros. It's up to the developer, currently, to take care that the _cutTrailingZeros_  prop's value isn't greater than _decimalScale_ value (if used) 

In case both _decimalScale_ and _cutTrailingZero_ props are used, if _cutTrailingZero_ > _decimalScale_  and _fixedDecimal_ is used, it will display cutTrailingZero amount of decimals instead of decimalScale   amount of decimals.

examples where this prop might be useful:
 say123,456.780 and cutTrailingZeros = 2 (because we do want to keep at least 2 decimals and cut all trailing zeros after 2nd decimal place) then the result will be => 123,456.78
but if 123,456.780 and  cutTrailingZeros = 3 then the result will be => 123,456.780 .

Github issue that this PR solves: [#611](https://github.com/s-yadav/react-number-format/issues/611)

Works in all browsers

### DIFF
--- a/example/src/index.js
+++ b/example/src/index.js
@@ -114,6 +114,17 @@ class App extends React.Component {
         </div>
 
         <div className="example">
+          <h3>Remove Trailing Zeros in Decimals</h3>
+          <NumberFormat
+            isNumericString
+            thousandSeparator
+            decimalScale={5}
+            fixedDecimalScale
+            cutTrailingZeros={2}
+          />
+        </div>
+
+        <div className="example">
           <h3>Format with pattern : Format credit card in an input</h3>
           <NumberFormat format="#### #### #### ####" mask="_" />
         </div>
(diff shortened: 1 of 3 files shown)


Yusuf007R:
👍

IGrobenskiTT:
> 👍

can this be merged to DEV?

Ostro:
i would really like to have this feature

abenhamdine:
so useful !

Issue #638, opened 2022-05-14
Controlled input can not be reset setting an falsy value


yeion7:
#### Describe the issue and the actual behavior

When you set an empty value to a controlled input the value is not reset

I tracked down this issue to `v4.4.2`, in `v.4.4.1` is working

https://github.com/s-yadav/react-number-format/compare/v4.4.1...v4.4.2

#### Describe the expected behavior

Input should be empty after set the new value

#### Provide a CodeSandbox link illustrating the issue

https://codesandbox.io/s/rhf-numberformat-defaultvalues-forked-7mtht0?file=/src/App.js

#### Provide steps to reproduce this issue

https://user-images.githubusercontent.com/9127504/168427830-1e3e4a87-e1c7-41c1-93cf-749631c9286c.mp4

#### Please check the browsers where the issue is seen
- [x] Chrome
- [ ] Chrome (Android)
- [ ] Safari (OSX)
- [ ] Safari (iOS)
- [ ] Firefox
- [ ] Firefox (Android)

antoantonydh:
@yeion7 I think, to reset you have to pass value as follows,

`setValue("money", "")`

Ref - https://github.com/s-yadav/react-number-format/issues/500#issuecomment-797342449

PR #639, opened 2022-05-14
add failing test for #638

yeion7:
Failing test to validate #638

### DIFF
--- a/test/library/input.spec.js
+++ b/test/library/input.spec.js
@@ -161,6 +161,40 @@ describe('NumberFormat as input', () => {
     expect(domInput.value).toEqual('$2,456,981');
   });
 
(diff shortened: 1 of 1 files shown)


test/library/input.spec.js
```diff
@@ -161,6 +161,40 @@ describe('NumberFormat as input', () => {
...
+    const input = wrapper.find('input');
+    const domInput = input.instance();
+
+    expect(domInput.value).toEqual('$1,000');
+
+    setButton.simulate('click')
+    expect(domInput.value).toEqual('$2,000');
+
+    resetButton.simulate('click')
+    expect(domInput.value).toEqual('');
```
yeion7, line 195: value is `$2,000` but should be `empty`

PR #648, opened 2022-06-29
fix: caret position when the first entered number is in the prefix

gyanta:
When `prefix` contains any number, and that number is typed as the first character, the caret position is calculated incorrectly.

In `getCaretPosition` the caret is positioned to the first occurrence of the typed character, which is in the prefix. `correctCaretPosition` then moves the caret to the first position after the prefix, but before the character we just typed in.

I tried to work it around by ensuring that the input always contains the prefix when `getCaretPosition` is called - not sure if this is the best way to fix it.

https://codepen.io/attila_n/pen/mdxdowd

- [X] Chrome
- [ ] Chrome (Android)
- [X] Safari (OSX)
- [ ] Safari (iOS)
- [ ] Firefox
- [ ] Firefox (Android)

### DIFF
--- a/src/number_format.js
+++ b/src/number_format.js
@@ -786,11 +786,17 @@ class NumberFormat extends React.Component {
 
   onChange(e: SyntheticInputEvent) {
     const el = e.target;
-    let inputValue = el.value;
     const { state, props } = this;
     const { isAllowed } = props;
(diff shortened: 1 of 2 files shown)


s-yadav:
I think this is already fixed on v5. Closing it.

PR #671, opened 2022-09-15
fix: decimal separator formatting

### LINKED ISSUE #670: 5.0.0-beta.4: bug - values returned by NumericFormat is multiplied when we use fixedDecimalScale
On latest beta version 5.0.0-beta.4 there is bug with multiply number without decimals when we pass fixedDecimalScale and decimalScale props. Check CodeSandbox for examples

Passing decimal scale = 2, fixedDecimalScale as true and value number 12 should display 12 not 1200.00

https://codesandbox.io/embed/charming-cookies-fxhpet

1. use NumericFormat from react-number-format
2. pass props with decimalScale set to 2 or 3
3. pass props fixedDecimalScale
4. Check differences between passing value as number and string

- [x] Chrome
- [ ] Chrome (Android)
- [ ] Safari (OSX)
- [ ] Safari (iOS)
- [ ] Firefox
- [ ] Firefox (Android)

davidskuza:
Fixes #670

### DIFF
--- a/src/utils.tsx
+++ b/src/utils.tsx
@@ -161,7 +161,7 @@ export function roundToPrecision(numStr: string, scale: number, fixedDecimalScal
   //if number is empty don't do anything return empty string
   if (['', '-'].indexOf(numStr) !== -1) return numStr;
 
-  const shoudHaveDecimalSeparator = numStr.indexOf('.') !== -1 && scale;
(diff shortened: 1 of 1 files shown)


s-yadav:
the check for fixedDecimalScale was missing for shouldHaveDecimalSeparator. Fixed on v5.0.0-beta.5 along with test.

https://github.com/s-yadav/react-number-format/commit/f1d6696200f5f85799d79bec980572a2cdcbd486

PR #676, opened 2022-09-28
Fix card expiry demo with usePatternFormat

nikhil-varma:
The card Expiry demo with the `usePatternFormat` hook does not trigger the `onValueChange` since the `NumberFormatBase` isn't getting it via the `...rest` variable.

### DIFF
--- a/documentation/v5/docs/customization.md
+++ b/documentation/v5/docs/customization.md
@@ -145,7 +145,7 @@ function CardExpiry(props) {
     return format(`${month}${year}`);
   };
 
-  return <NumberFormatBase format={_format} {...rest} />;
+  return <NumberFormatBase format={_format} {...props} />;
 }
 ```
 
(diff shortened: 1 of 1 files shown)


documentation/v5/docs/customization.md
```diff
@@ -145,20 +145,20 @@ function CardExpiry(props) {
     return format(`${month}${year}`);
   };
 
-  return <NumberFormatBase format={_format} {...rest} />;
+  return <NumberFormatBase format={_format} {...props} />;
```
s-yadav, line 148: Ohh, we need to spread both the props coming from parent and the return of usePatternFormat. Something like this.
```suggestion
  return <NumberFormatBase  {...props} format={_format} {...rest}/>;
```

nikhil-varma: Done!

s-yadav:
Now we are passing down all the props, so this fix is no longer required.

PR #703, opened 2022-11-03
Pass type parameter to SyntheticEvent

yukiyokotani:
Currently the type of `event` that `SourceInfo` has is `SyntheticEvent` with no type argument. In the implementation, I believe that it is the `input` element that fires the event, so `SyntheticEvent<HTMLInputElement>` should be more appropriate here. This change will improve integration with other libraries such as MUI in terms of TypeScript types.

https://github.com/s-yadav/react-number-format/blob/master/src/types.ts#L23
befor: `SyntheticEvent`
after: `SyntheticEvent<HTMLInputElement>`

- [x] Chrome
- [ ] Chrome (Android)
- [ ] Safari (OSX)
- [ ] Safari (iOS)
- [ ] Firefox
- [ ] Firefox (Android)

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -20,7 +20,7 @@ export enum SourceType {
 }
 
 export interface SourceInfo {
-  event?: SyntheticEvent;
+  event?: SyntheticEvent<HTMLInputElement>;
   source: SourceType;
 }
 


s-yadav:
Make sense. Thanks for the PR.

s-yadav (approved):

PR #713, opened 2022-12-20
Mark module as side effect free

topaxi:
This should enable webpack and other bundlers to only import used functions and components.


s-yadav:
Thanks for the PR.

Issue #725, opened 2023-01-31
Using decimalSeparator with a suffix containing more than one character brakes the seperator behavior


musbou:
#### Describe the issue and the actual behavior
As stated in title, when using decimalSeparator in conjunction of a suffix with more than one character, typing the default separator `.` (or any other separator if you have a list allowedDecimalSeparator) does not move the caret. It will leave the cursor before the dot and not after. So you'll be inputing whole numbers still, unless you use the mouse or arrow keys to navigate after the separator.

Having one character suffix fixes this issue.

Example of usage:
```javascript
<NumericFormat
     value={0}
     decimalSeparator=","
     allowedDecimalSeparators={["%"]}
     decimalScale={2}
     suffix=" €"
/>
```
#### Describe the expected behavior
Inputing the decimal separate should leave the cursor after the separator.

#### Provide a CodeSandbox link illustrating the issue
https://codesandbox.io/s/tender-dream-914v43

#### Provide steps to reproduce this issue
See codesandbox.

#### Please check the browsers where the issue is seen
- [x] Chrome
- [x] Firefox

musbou:
After some experiments, it turns out that the presence of more than 1 character is the issue. I update the issue.

romanopassalacqua:
I'm also facing this issue.

I did some debugging and the `if` in [src/utils.tsx#L334](https://github.com/s-yadav/react-number-format/blob/master/src/utils.tsx#L334) compares the inputted char (key pressed) with the formatted char. Should this `if` check for `allowedDecimalSeparators` and `decimalSeparator`? I think it has something to do with this `if` because if the key pressed is the `decimalSeparator` it works.

#### [src/utils.tsx#L334](https://github.com/s-yadav/react-number-format/blob/master/src/utils.tsx#L334)
```js
  for (let i = 0; i < curValLn; i++) {
    indexMap[i] = -1;
    for (let j = 0, jLn = formattedValueLn; j < jLn; j++) {
      // ↓
      if (curValue[i] === newFormattedValue[j] && addedIndexMap[j] !== true) {
        indexMap[i] = j;
        addedIndexMap[j] = true;
        break;
      }
    }
  }
```
---
Not sure if #673 could be related to this

sormpe:
Hi,

I faces this too, and opened a PR to fix it. [#743](https://github.com/s-yadav/react-number-format/issues/743)

sormpe:
[...]

PR #743, opened 2023-03-08
Fix decimalSeparator with a suffix containing more than one character

sormpe:

This PR fixes issue described in [#725](https://github.com/s-yadav/react-number-format/issues/725)

As [romanopassalacqua](https://github.com/romanopassalacqua) suggested, we now check if char is in allowedDecimalSeparators in getCaretPosition. Before we only checked if chars were the same resulting that indexMap was not affected. This caused problem described in [#725](https://github.com/s-yadav/react-number-format/issues/725) when using some of the allowedDecimalSeparators which is not (default) decimalSeparator.

 [#725](https://github.com/s-yadav/react-number-format/issues/725)

```
        <div className="example">
          <h3>A suffix containing more than one character</h3>
          <NumericFormat
            suffix={' €'}
            allowedDecimalSeparators={['.', ',']}
          />
        </div>

        <div className="example">
          <h3>A long suffix</h3>
          <NumericFormat
            suffix={'longsuffix'}
            allowedDecimalSeparators={['.', ',']}
          />
        </div>

        <div className="example">
          <h3>A superlong suffix</h3>
          <NumericFormat
            suffix={'superlongsuffix with spaces'}
            allowedDecimalSeparators={['.', ',']}
          />
        </div>

```
#### Video

https://user-images.githubusercontent.com/19435681/223791397-3c7efa76-29e7-4f96-846b-933a2813d542.mp4

- [x] Chrome
- [ ] Chrome (Android)
- [ ] Safari (OSX)
- [ ] Safari (iOS)
- [x] Firefox
- [ ] Firefox (Android)

### DIFF
--- a/src/numeric_format.tsx
+++ b/src/numeric_format.tsx
@@ -473,6 +473,11 @@ export function useNumericFormat<BaseType = InputAttributes>(
 
   const isValidInputCharacter = (inputChar: string) => {
     if (inputChar === decimalSeparator) return true;
+
+    if (allowedDecimalSeparators.includes(inputChar)) {
+      return true;
+    }
+
     return charIsNumber(inputChar);
   };
 
--- a/src/numeric_format.tsx
+++ b/src/numeric_format.tsx
@@ -474,7 +474,7 @@ export function useNumericFormat<BaseType = InputAttributes>(
   const isValidInputCharacter = (inputChar: string) => {
     if (inputChar === decimalSeparator) return true;
 
-    if (allowedDecimalSeparators.includes(inputChar)) {
+    if (allowedDecimalSeparators?.includes(inputChar)) {
       return true;
     }
 
--- a/src/utils.tsx
+++ b/src/utils.tsx
@@ -331,7 +331,13 @@ export function getCaretPosition(
   for (let i = 0; i < curValLn; i++) {
     indexMap[i] = -1;
     for (let j = 0, jLn = formattedValueLn; j < jLn; j++) {
-      if (curValue[i] === newFormattedValue[j] && addedIndexMap[j] !== true) {
+      const isAllowedDecimalSeparator =
+        isValidInputCharacter(curValue[i]) && isValidInputCharacter(newFormattedValue[j]);
+
+      if (
+        (curValue[i] === newFormattedValue[j] || isAllowedDecimalSeparator) &&
+        addedIndexMap[j] !== true
+      ) {
         indexMap[i] = j;
         addedIndexMap[j] = true;
         break;
--- a/src/utils.tsx
+++ b/src/utils.tsx
@@ -333,10 +333,9 @@ export function getCaretPosition(
     for (let j = 0, jLn = formattedValueLn; j < jLn; j++) {
       const isAllowedDecimalSeparator =
         isValidInputCharacter(curValue[i]) && isValidInputCharacter(newFormattedValue[j]);
-
       if (
-        (curValue[i] === newFormattedValue[j] || isAllowedDecimalSeparator) &&
-        addedIndexMap[j] !== true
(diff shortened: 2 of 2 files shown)


sormpe:
Apparently this change break some tests, could you [s-yadav](https://github.com/s-yadav) check this out?

sormpe:
I noticed that my fix works only if decimal separator is the first char we type into. In other cases startIndex should be one greater. We can check if allowed decimal separator was used by adding boolean flag, something like this:

```
export function getCaretPosition(
  newFormattedValue: string,
  lastFormattedValue: string,
  curValue: string,
  curCaretPos: number,
  boundary: boolean[],
  isValidInputCharacter: (char: string) => boolean,
) {
  /**
   * if something got inserted on empty value, add the formatted character before the current value,
   * This is to avoid the case where typed character is present on format characters
   */

  const firstAllowedPosition = boundary.findIndex((b) => b);
  const prefixFormat = newFormattedValue.slice(0, firstAllowedPosition);
  if (!lastFormattedValue && !curValue.startsWith(prefixFormat)) {
    curValue = prefixFormat + curValue;
    curCaretPos = curCaretPos + prefixFormat.length;
  }

  const curValLn = curValue.length;
  const formattedValueLn = newFormattedValue.length;

  // create index map
  const addedIndexMap: { [key: number]: boolean } = {};
  const indexMap = new Array(curValLn);

  let wasAllowedDecimalSeparatorUsed = false;

  for (let i = 0; i < curValLn; i++) {
    indexMap[i] = -1;
    for (let j = 0, jLn = formattedValueLn; j < jLn; j++) {
      const isAllowedDecimalSeparator =
        isValidInputCharacter(curValue[i]) && isValidInputCharacter(newFormattedValue[j]);
      if (
        (curValue[i] === newFormattedValue[j] && addedIndexMap[j] !== true) ||
        (isAllowedDecimalSeparator && i === 0 && j === 0)
      ) {
        indexMap[i] = j;
        addedIndexMap[j] = true;
        // ADD FLAG HERE!
        wasAllowedDecimalSeparatorUsed = true;
        break;
      }
    }
  }

  /**
   * For current caret position find closest characters (left and right side)
   * which are properly mapped to formatted value.
   * The idea is that the new caret position will exist always in the boundary of
   * that mapped index
   */
  let pos = curCaretPos;
  while (pos < curValLn && (indexMap[pos] === -1 || !isValidInputCharacter(curValue[pos]))) {
    pos++;
  }

  // if the caret position is on last keep the endIndex as last for formatted value
  const endIndex = pos === curValLn || indexMap[pos] === -1 ? formattedValueLn : indexMap[pos];

  pos = curCaretPos - 1;
  while (pos > 0 && indexMap[pos] === -1) pos--;
  const startIndex = pos === -1 || indexMap[pos] === -1 ? 0 : indexMap[pos] + 1;

  /**
   * case where a char is added on suffix and removed from middle, example 2sq345 becoming $2,345 sq
   * there is still a mapping but the order of start index and end index is changed
   */
  if (startIndex > endIndex) return endIndex;

  /**
   * given the current caret position if it closer to startIndex
   * keep the new caret position on start index or keep it closer to endIndex
   */

// USE FLAG HERE!
 if (wasAllowedDecimalSeparatorUsed) {
    return curCaretPos - startIndex < endIndex - curCaretPos ? startIndex + 1 : endIndex;
  }

  return curCaretPos - startIndex < endIndex - curCaretPos ? startIndex : endIndex;
}
```

This, however, break the tests, and I'm unsure if it break funtionality in some other cases.

s-yadav:
Hey @sormpe Thanks for PR. couldn't look into this earlier, will check this.

s-yadav:
As this was incorporated with some other fix. closing the PR.

PR #750, opened 2023-04-12
Remove unnecessary useEffect hook

aashu16:
Part of the logic in the NumberFormatBase component that is responsible for handling prop changes does not need to be wrapped in a useEffect hook. Moving this logic outside the hook also ensures that it runs before the component is rendered, preventing unnecessary re-renders.

This PR does the following:
* Move the logic outside the useEffect hook.
* Move the affected code so that it follows the declaration of any values/functions that it uses/calls.

### DIFF
--- a/src/number_format_base.tsx
+++ b/src/number_format_base.tsx
@@ -69,28 +69,6 @@ export default function NumberFormatBase<BaseType = InputAttributes>(
     onFormattedValueChange(values, source);
   };
 
-  // check if there is any change in the value due to props change
-  useEffect(() => {
-    const newFormattedValue = (format as FormatInputValueFunction)(numAsString);
-
-    // if the formatted value is not synced to parent, or if the formatted value is different
-    if (lastUpdatedValue.current === undefined || newFormattedValue !== lastUpdatedValue.current) {
-      const input = focusedElm.current;
-
-      // formatting can remove some of the number chars, so we need to fine number string again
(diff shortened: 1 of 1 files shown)


s-yadav:
useEffect here is intentional, as you can see on failing test. 
There are two reasons,
1) As this takes care of informing the parent about formatted value change on prop change. Changing parent state while rendering children is invalid and can cause unexpected bugs.

2) Also, the updateValue depends on the internal state and the internal state change is only reflected after the react bails out and try to render again, but the logic inside component execution doesn't stop, so it can call onValueChange callback twice. Using useEffect here make sure, it happens after the state is stable and applied to dom.

Take this example.
- props changed due to change in parent.
- number format tries to update internal state based on prop.
- updateValue gets called with previous state and new props
- internal state updates
- updateValue gets called with new state and new props

s-yadav:
However this can be optimized with some refactor. to avoid 2 renders within react number format component. But we still have to inform the parent with onValueChange on useEffect.

aashu16:
Been thinking about the best way to tackle this. I will try and push an update before the end of the week.

s-yadav:
@aashu16 I have fixed this on ongoing fixes. yet to merge and release.

aashu16:
Oh cool. I already started working on a fix yesterday, but no worries!

Have you pushed those changes to a remote branch yet? Just wondering if I could take a look at the approach you took.

PR #752, opened 2023-04-21
Add support for minDecimalScale

RayzorFlash:

- In many currency-related use cases, a minimal decimal precision of 2 digits is common. This way, large numbers get displayed with two padded zeroes (i.e. $1,000,000.00) while small numbers support added precision of up-to n digits ($0.89765412 if n=8)

- This functionality is not currently possible with this library due to fixedDecimalScale enforcing EXACTLY n digits of precision, while decimalScale provides only a MAXIMUM number of digits when used on its own

- In a future release (since it would be a breaking change) we can probably remove `fixedDecimalScale` and use only `minDecimalScale` and `decimalScale` to clamp the bounds of precision and satisfy every use case

```
<NumberFormat minDecimalScale={2} decimalScale={8} value={24}/> 
```

- [X] Chrome
- [ ] Chrome (Android)
- [ ] Safari (OSX)
- [ ] Safari (iOS)
- [ ] Firefox
- [ ] Firefox (Android)

### DIFF
--- a/documentation/v5/docs/numeric_format.md
+++ b/documentation/v5/docs/numeric_format.md
@@ -130,6 +130,31 @@ import { NumericFormat } from 'react-number-format';
    ></iframe>
 </details>
 
+### minDecimalScale `number`
+
+**default**: `undefined`
+
+If defined, it enforces a minimum number of digits after the decimal point.
+
+```js
+import { NumericFormat } from 'react-number-format';
+
+<NumericFormat value={12323.3334} minDecimalScale={3} />;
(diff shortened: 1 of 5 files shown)


s-yadav:
@RayzorFlash Thanks for the PR.
what about we introduce two new props. `minimumFractionDigits` and `maximumFractionDigits`. This aligns with Intl.NumberFormat options.

And then we can mark decimal scale and fixedDecimalScale deprecated which we can remove on next major release.

PR #763, opened 2023-05-25
Migrate tests

aashu16:
The primary goals of this PR are:
- Replace Karma/Jasmine with Vitest
- Migrate tests from Enzyme to @testing-library/react

Checklist:
- [x] Configure vitest
- [x] Update test utils
- [x] Update test files
   - [x] input_numeric_format
   - [x] keypress_and_caret
   - [x] utils
   - [x] input
   - [x] format_as_text
- [x] Remove packages that are no longer required - Karma, Jasmine, Enzyme + related dependencies

PR #779, opened 2023-07-22
Test upgrade 

s-yadav:
- [ ] Chrome
- [ ] Chrome (Android)
- [ ] Safari (OSX)
- [ ] Safari (iOS)
- [ ] Firefox
- [ ] Firefox (Android)


test/library/input.spec.jsx
```diff
@@ -0,0 +1,1076 @@
...
+        <NumericFormat
+          decimalScale={2}
+          prefix="$"
+          thousandSeparator
+          value="100.10"
+          valueIsNumericString
+        />,
+      );
+
+      expect(input).toHaveValue('$100.10');
```
s-yadav, line 1054: **Suggestion**
When we rerendering so many times, we can create in place component/utils, to reduce the duplicats.
```suggestion
     
     const CustomNumericFormat = (props) => (
       <NumericFormat
          decimalScale={2}
          prefix="$"
          thousandSeparator
          valueIsNumericString
          {...props}
        />
     )

      const { input, user, rerender } = await render( <CustomNumericFormat value="100.0"/>);

      expect(input).toHaveValue('$100.0');

      rerender(<CustomNumericFormat  value="123.00"  />);

      expect(input).toHaveValue('$123.00');

      rerender(<CustomNumericFormat value="132.000" />);

      expect(input).toHaveValue('$132.00');

      rerender(<CustomNumericFormat value="100.10" />);

      expect(input).toHaveValue('$100.10');
```

aashu16: The plan here is to split tests like this into smaller tests, but point taken.

test/library/format_as_text.spec.jsx
```diff
@@ -1,51 +1,60 @@
...
 /*** format_number input as text ****/
 describe('NumberFormat as text', () => {
   it('should format numbers to currency', () => {
-    const wrapper = mount(
+    rtlRender(
       <NumericFormat value={2456981} displayType={'text'} thousandSeparator={true} prefix={'$'} />,
     );
-    expect(wrapper.find('span').text()).toEqual('$2,456,981');
+
+    const span = screen.getByTestId('rnf-renderText-span');
```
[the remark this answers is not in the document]
aashu16: I don't think all the changes merged in #763 are ready to be merged into main yet. I think it would be better to close this PR and create a new one when some of the other ongoing work in the dev branch has been completed.

aashu16:
The changes from #763 are not ready to be merged into main yet. I think it would be better to close this PR and create a new one when some of the other ongoing work in the dev branch has been completed.

test/library/input.spec.jsx
```diff
@@ -0,0 +1,1076 @@
...
+          className="foo"
+          renderText={(formattedValue, props) => (
+            <span data-testid="input-renderText-span" {...props}>
+              {formattedValue}
+            </span>
+          )}
+        />,
+      );
+
+      const span = screen.getByTestId('input-renderText-span');
```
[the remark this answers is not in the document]
aashu16: The span here exists inside this test. It's not a part of the library.

test/library/input.spec.jsx
```diff
@@ -0,0 +1,1076 @@
...
+    it.skip('should always add inputMode numeric to pattern format, even for Iphone/IPad device', async () => {
+      navigator['__defineGetter__']('platform', async () => {
+        return 'iPhone';
+      });
+      const { input, rerender } = await render(<PatternFormat />);
+      expect(input.getAttribute('inputmode')).toBe('numeric');
+
+      rerender(<PatternFormat format="##" />);
+      expect(input.getAttribute('inputmode')).toBe('numeric');
+    });
```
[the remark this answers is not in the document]
aashu16: Just trying to understand the purpose behind checking for a component attribute before it is mounted. The component doesn't "exist" yet.

aashu16:
A couple of small changes left to make - then this should be ready to merge.

@s-yadav take a look when you can and let me know if you have any further comments.

I would rather create a new release branch once the remaining changes have been made and then merge that into main. So, I am going to close this PR now. Please feel free to reopen if you would rather merge this directly.

s-yadav:
@aashu16, 
Aaah better to use the same PR, so I can just check on changes that has been done after the last review. Or else I have to go through all the changes again. Not sure if we reopen github will allow seeing the changes made afterwards.

Anyways other option is that I can check from commit ranges. But Yes please don't squash all the commits until the review is finished.

aashu16:
> @aashu16, Aaah better to use the same PR, so I can just check on changes that has been done after the last review. Or else I have to go through all the changes again. Not sure if we reopen github will allow seeing the changes made afterwards.

Sure, no problem. I will leave this open for now and keep making changes. You should still be able to see your comments and my responses. I marked them as resolved, but feel free to undo that.

> Anyways other option is that I can check from commit ranges. But Yes please don't squash all the commits until the review is finished.

Not planning on squashing anything 😄

test/test_util.js
```diff
@@ -1,213 +1,109 @@
...
-      newValue =
-        selectionStart !== selectionEnd
-          ? currentValue.substring(0, selectionStart) +
-            key +
-            currentValue.substring(selectionEnd, currentValue.length)
-          : currentValue.substring(0, selectionStart) +
-            key +
-            currentValue.substring(selectionStart, currentValue.length);
-    }
+  const specialKeys = ['{Backspace}', '{Delete}', '{ArrowLeft}', '{ArrowRight}'];
```
s-yadav, line 46: Q: Why do we need manual handling for specialKeys? Any issue with testing library? If yes then can we add it on comment.

test/test_util.js
```diff
@@ -1,213 +1,109 @@
...
+    } else {
+      let newValue;
 
-export function simulateNativeMouseUpEvent(input, selectionStart) {
-  input.setSelectionRange(selectionStart, selectionStart);
-  userEvent.click(input);
+      newValue = v.slice(0, start) + key + v.slice(end, v.length);
+      fireEvent.change(input, { target: { value: newValue } });
+      end = start;
+    }
```
s-yadav, line 82: Q: Why do need custom handling here?
Won't just this work.
```
user.type(input, key, { initialSelectionStart: start, initialSelectionEnd: end });
```

Or
```
input.setSelectionRange(selectionStart, selectionEnd);
userEvent.type(input, key);
```

test/test_util.js
```diff
@@ -1,213 +1,109 @@
...
-      selectionStart,
-      selectionEnd,
-      setSelectionRange: setSelectionRange || noop,
-      focus: noop,
-    },
-  );
-
-  input.simulate('focus', focusEvent);
+export function simulateFocusEvent(input) {
+  input.focus();
```
s-yadav, line 95: **Suggestion/Test Bug**
So there are three cases how focus can happen.
- Due to mouse click on specific caret position. In which case the caret is placed at specific position. Which I think the usage is replaced now with simulateMousUpEvent.
- Due to tab press, in which case whole input value gets selected.
- Manual input.focus call, in which case also whole input value gets selected.

If we catering all the three cases with this util, we should have selectionStart and selectionEnd as param.

In case we are catering 2nd and 3rd, we need to manually set the selectionStart to 0 and selectionEnd at input.value.length

Reason: input.focus() in JSDom puts the caret position at the end, which is not same as browser.

test/test_util.js
```diff
@@ -1,213 +1,109 @@
...
 }
 
-export async function wait(delay) {
-  return new Promise((resolve) => {
-    setTimeout(() => {
-      resolve();
-    }, delay);
-  });
+export async function simulatePaste(user, input, data) {
+  input.focus();
```
s-yadav, line 107: **Suggestion/Possible bug**
JSDOM puts the caret position at the end, so paste will always happen at the end. Ideally, we should take selectionStart and selectionEnd as params, where selectionStart is defaulted to 0 and selectionEnd is defaulted to input.value.length.
And then set that selection after doing input.Focus().

This way we can also handle pasting at specific position, or selecting partial text and pasting.

s-yadav:
@aashu16 Merged master and fixed some broken tests which was incorrect on master as well.

test/test_util.js
aashu16: That is the plan going forward. As it stands now, the tests that require pasting don't need a selectionRange to be set. So, it would be an addition with no effect.

I do have some tests in mind which could use this suggested feature (don't want to delay this PR further with entirely new testing scenarios at this stage). Personally think it would be better to add this when there are tests that ready to use it.

test/test_util.js
s-yadav: Actually, I tried `await user.type(input, key, { initialSelectionStart: start, initialSelectionEnd: end });`, looks like in some cases it's not working, basically due to JSDOM, the selected text is not getting cleared first. Need to dig more, but for now we can keep what you have added.

test/test_util.js
s-yadav: Yes checked currently we are clearing the input first and then doing paste in the test case. I think its ok to keep our util ready for that, as its just small change. Anyways never mind, I have already added that.

test/test_util.js
aashu16: Haven't pushed the relevant commits yet but I have already fixed this locally.

test/test_util.js
aashu16: It's not the special keys but all other keys that need conditional handling (the next few lines on code should probably make this clear).

I have not faced this issue with testing-library elsewhere but cannot say with certainty what the root cause is. Consider this a workaround for now. When time permits, I will try and come back to this and dig deeper.

I will add a brief comment to explain this.

s-yadav: Yes, most likely its issue with jsdom + testing lib.

test/test_util.js
aashu16: Just looked at the change you just pushed to `simulatePaste`. I am not sure about default values `selectionStart` and `selectionEnd`. In `simulateKeyInput`, if `selectionEnd` is not provided, it is set to `selectionStart`.

IMO, it would make sense for these two functions to follow the same logic, given what they do is fairly similar.

s-yadav: Ohh, I simulated the input.focus from browser as default. But make sense, we should have same default as simulateKeyInput. Fell free to change, or else i will update it.

aashu16: Leave this with me. I will be working on this later today.

aashu16:
@s-yadav, I think this is just about ready to merge now. Let me know if you have any comments regarding the recent commits.

(Please don't merge yet.)

PR #794, opened 2023-09-09
Add unit tests

aashu16:
Adding tests for:

- [x] applyThousandSeparator
- [x] roundToPrecision
- [x] toNumericString
- [x] splitDecimal
- [x] limitToScale
- [x] fixLeadingZero
- [x] findChangedIndex
- [x] findChangeRange
- [x] getCaretBoundary (numeric_format)
- [x] getCaretBoundary (pattern_format)
- [x] getCaretPosInBoundary
- [x] getCaretPosition


aashu16:
@s-yadav, I would like your opinion on the tradeoffs associated with the following 2 approaches.

In their current form, the tests in `test/library/utils` are very explicit in how the tests are run and type definitions are easy to grok. However, every new function that's added to the test suite requires a few lines of boilerplate code.

(There are some easy opportunities for refactoring - like sharing a common TestType definition and some sensible renaming - but that's not important here.)

Here's an example from `test/library/utils/split-decimal.spec.ts`.
```typescript
type ReturnType = {
  beforeDecimal: string;
  afterDecimal: string;
  hasNegation: boolean;
  addNegation: boolean;
};

type CaseType = {
  numStr: string;
  allowNegative: boolean;
  expected: ReturnType;
};

type TestType = { label: string; cases: CaseType[] };

const testCases: TestType[] = [
  {
    label: 'allowNegative = true',
    cases: [
      {
        numStr: '100.40',
        allowNegative: true,
        expected: {
          beforeDecimal: '100',
          afterDecimal: '40',
          hasNegation: false,
          addNegation: false,
        },
      },
      ...
    ],
  },
  ...
];

for (const testCase of testCases) {
  describe(testCase.label, () => {
    test.each(testCase.cases)('$numStr', ({ numStr, allowNegative, expected }) => {
      expect(splitDecimal(numStr, allowNegative)).toStrictEqual(expected);
    });
  });
}
```

The following is an implementation which does away with the boilerplate but also loses some explicitness along with, maybe, a little flexibility. But to add tests for a new function, all one needs to do is create a "table" of tests and pass it to `runTableTests`.

Thoughts?

```typescript
// ReturnType same as above.

// Arguments to splitDecimal are (numStr, allowNegative)
type Arguments = [string, boolean];

const testCases: TestType<Arguments, ReturnType>[] = [
  [
    'allowNegative = true',
    [
      [
        '100.40',
        true,
        {
          beforeDecimal: '100',
          afterDecimal: '40',
          hasNegation: false,
          addNegation: false,
        },
      ],
      ...
    ],
  ],
  ...
];

runTableTests(testCases);
```

Declared elsewhere:
```typescript
type CaseType<A extends unknown[], R extends unknown> = [...arguments: A, expected: R];

export type TestType<T extends unknown[], U extends unknown> = [label: string, cases: CaseType<T, U>[]];

export const runTableTests = <T extends unknown[], U extends unknown, C extends CaseType<T, U>>(
  testCases: TestType<C>[],
) => {
  for (const testCase of testCases) {
    describe(testCase[0], () => {
      test.each(testCase[1])('%s', (...args) => {
        expect(splitDecimal(...(args.slice(0, args.length - 1) as Arguments))).toStrictEqual(
          args[args.length - 1],
        );
      });
    });
  }
};
```

s-yadav:
@aashu16 Yes for utils (unit-test) we can have this abstraction. As they are always like pass some input, and expect some output.
One minor suggestion on the config object, they can follow the following type.
```js
type TestCases<Input, Output> = {
  label: string,
  cases: {
    input: Input,
    output: Output
  }[] 
}[]
```
Basically, just have the input and output explicit key defined on the config you are suggestion.

Also, I assume we want to follow the config pattern only for unit tests for the utils. For any behaviour level test, the generic config pattern doesn't do well, as you mentioned it loses the explicitness of the test and flexibility.

aashu16:
Sorry if I missed something in your response, but what do you think about the tuple based approach?

```typescript
// Details in my previous comment.

type TestType = { label: string; cases: CaseType[] };
// vs
type TestType<T extends unknown[], U extends unknown> = [label: string, cases: CaseType<T, U>[]];
```

> Also, I assume we want to follow the config pattern only for unit tests for the utils. For any behaviour level test, the generic config pattern doesn't do well, as you mentioned it loses the explicitness of the test and flexibility.

Yes, that's right. We can revisit this if a pattern emerges in the future but I can't see one now and it's not a priority anyway.

aashu16:
After thinking about this a little more, I have come to the conclusion that the first approach (using objects instead of tuples) is the correct one. The explicitness is worth the minor repetition.

s-yadav:
Yes make sense. The object is always readable.

PR #801, opened 2023-09-25
Type `type` prop as `string`

### LINKED ISSUE #799: The `type` prop should accept `string` as indicated in documentation

There is a disconnect between the [documentation for the `type` prop](https://s-yadav.github.io/react-number-format/docs/props/#type-string), which states that it accepts any `string`, and the actual TypeScript types, which narrow the accepted values to "text", "tel", or "password":

https://github.com/s-yadav/react-number-format/blob/e2179f9e9d2c4ddd1ab714ff385412b53e9ed3e7/src/types.ts#L68-L69

Since `string` is the correct type for the `type` property of HTML5 `<input>` elements, and browsers fall back to the default behavior of `type="text"` when the type is outside the list of supported values, I believe the `type` prop should be typed as `type?: string`.

In my testing, all `string`s work as expected during runtime. I have a [branch](https://github.com/jakeboone02/react-number-format/tree/type-prop-string) ready for PR if this change would be accepted.

https://codesandbox.io/s/type-demo-forked-p4lh8w?file=/src/App.js

- Use prop `type="number"` in a TypeScript project.

_Not specific to a browser; issue is with TypeScript types._

Original issue: mantinedev/mantine#4821.

jakeboone02:
Fixes #799. Issue text replicated below:
___

There is a disconnect between the [documentation for the `type` prop](https://s-yadav.github.io/react-number-format/docs/props/#type-string), which states that it accepts any `string`, and the actual TypeScript types, which narrow the accepted values to "text", "tel", or "password":

https://github.com/s-yadav/react-number-format/blob/e2179f9e9d2c4ddd1ab714ff385412b53e9ed3e7/src/types.ts#L68-L69

Since `string` is the correct type for the `type` property of HTML5 `<input>` elements, and browsers fall back to the default behavior of `type="text"` when the type is outside the list of supported values, I believe the `type` prop should be typed as `type?: string`.

In my testing, all `string`s work as expected during runtime.

https://codesandbox.io/s/type-demo-forked-p4lh8w?file=/src/App.js

- Use prop `type="number"` in a TypeScript project.

_Not specific to a browser; issue is with TypeScript types._

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -66,7 +66,7 @@ export type IsCharacterSame = (compareProps: {
 }) => boolean;
 
 type NumberFormatBase = {
-  type?: 'text' | 'tel' | 'password';
+  type?: string;
   displayType?: 'input' | 'text';
   inputMode?: InputAttributes['inputMode'];
   renderText?: (formattedValue: string, otherProps: Partial<NumberFormatBase>) => React.ReactNode;


s-yadav:
https://github.com/s-yadav/react-number-format/issues/799#issuecomment-1742145157

jakeboone02:
Understood. Thanks for reviewing.

PR #813, opened 2023-11-22
refactor: clean up `updateValueAndCaretPosition`

csantos1113:
#### Describe the change

- `updateValueAndCaretPosition` is used only once
- its params `setCaretPosition` and `caretPos` were always set to something static
  - `setCaretPosition: true`
  - `caretPos` not passed, hence `undefined`

This PR removes these 2 params; and removes unnecessary checks

**Note:** not sure if these params were used for different purposes in the past, or if they are planned to used differently in the future, but they are not used as of right now, so I'm removing them.


s-yadav:
Yes, this method was earlier used when the value is changing from props, but that logic is changed now. 
Your changes look great. Will merge. Thanks

s-yadav (approved):

PR #814, opened 2023-11-22
fix: `setCaretTimeout` checks

### LINKED ISSUE #811: `suffix` not working properly in `jest` + `testing-library`

typing a number with 2 digits in a NumericFormat with `suffix` result in the wrong value when done in a jest test.

typing with jest should result in the same behavior as if typing in browser

https://codesandbox.io/s/recursing-poitras-rxtjkj?file=/src/index.test.js

- go to "Browser" in CodeSandbox
- type `9`
- notice:
  - input value is `9 people` ✅ 
  - caret position is after the `9` ✅ 
- type `1`
- notice:
  - input value is `91 people` ✅ 
  - caret position is after the `1` ✅ 
- go to "Tests" in CodeSandbox
  - the test will type `91` and expect `91 people` as outcome
  - the input value ends up being `9 people` 😢🔴

- [ ] Chrome
- [ ] Chrome (Android)
- [ ] Safari (OSX)
- [ ] Safari (iOS)
- [ ] Firefox
- [ ] Firefox (Android)
- [x] `jest` + `testing-library`

#### Additional details

After a lot of local debugging; it seems that when [`userEvent`](https://testing-library.com/docs/user-event/utility#type) tries to type `1` (from `91`); the caret is placed at the end of the input, so the input value is: `9 people1` which gets cleaned up back to `9 people` making it fail

csantos1113:
## Describe the change

Fixes #811

I'm opening this PR as an attempt to resolve `jest` + `testing-library` issue reported in #811.

I have an open question on that issue https://github.com/s-yadav/react-number-format/issues/811#issuecomment-1819527005 where I wondered why `el.selectionStart !== el.selectionEnd` existed; but after a lot of thinking I cannot see a reason for it

### My thinking

`selectionStart` and `selectionEnd` will always be the same when the caret is placed and nothing is selected:

https://github.com/s-yadav/react-number-format/assets/9648559/fb3c2791-9e43-40c8-b358-91d4a2cb85ac

so `el.selectionStart !== el.selectionEnd` will ***only*** happen if the user have text selected:

https://github.com/s-yadav/react-number-format/assets/9648559/9ff9066c-de30-4c83-ba8f-a1b2bd6fb690

which differs from the intention described by the inline comment that is in the code:

```
    /* setting caret position within timeout of 0ms is required for mobile chrome,
    otherwise browser resets the caret position after we set it
    We are also setting it without timeout so that in normal browser we don't see the flickering */
```

But please let me know if I've missed something else on my analysis 🙏

### DIFF
--- a/src/number_format_base.tsx
+++ b/src/number_format_base.tsx
@@ -116,7 +116,7 @@ export default function NumberFormatBase<BaseType = InputAttributes>(
     setCaretPosition(el, caretPos);
 
     timeout.current.setCaretTimeout = setTimeout(() => {
-      if (el.value === currentValue && el.selectionStart !== el.selectionEnd) {
+      if (el.value === currentValue && el.selectionStart !== caretPos) {
         setCaretPosition(el, caretPos);
       }
     }, 0);


csantos1113:
Hey @s-yadav - I'm gently pinging you for when you get a chance to review this PR. Thank you!

csantos1113:
@s-yadav 👋  when you get a chance, this would be much appreciated!

s-yadav:
Yes, the selectionStart !== selectionEnd mostly looks unintentional. the check with caretPos, is correct. 
I will see if it can be accompanied by a test as this was mostly a side-effect of not having a test.

s-yadav (approved):

PR #815, opened 2023-11-23
fix: pass currentTarget to onFocus

csantos1113:

This change is based on https://stackoverflow.com/a/66086044 `currentTarget`:  only exists if used right from the `onFocus` event; but we get currentTarget **null** because `onFocus(e)` is called **within** a timeout.

References:
- https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget#currenttarget_versus_target
- https://github.com/adobe/react-spectrum/blob/9d8c74a344c9b7539f00fa58d0a02b062b6e830e/packages/%40react-aria/interactions/src/useFocus.ts#L64

### DIFF
--- a/src/number_format_base.tsx
+++ b/src/number_format_base.tsx
@@ -383,6 +383,7 @@ export default function NumberFormatBase<BaseType = InputAttributes>(
     if (e.persist) e.persist();
 
     const el = e.target;
+    const currentTarget = e.currentTarget;
     focusedElm.current = el;
 
     timeout.current.focusTimeout = setTimeout(() => {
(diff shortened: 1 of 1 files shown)


sagar-girhepunje (approved):

csantos1113:
@s-yadav 👋  when you get a chance, this would be much appreciated!

s-yadav:
Hey @csantos1113 , was little occupied with some personal things for last 4 months, so didn't check. 
The changes looks good. Please do accompany the fix with a test.

s-yadav:
Never mind I added test, and merging this.

s-yadav (approved):

PR #819, opened 2023-12-29
fix: Optionally chain the `setCaretPosition` element focus

Triggs125:
We are using a custom input component which utilizes increment/decrement buttons to add/subtract a set step to the value. When either button is clicked on the element which is passed to the `setCaretPosition` function does not have a focus function so it errors out (see screenshot below) and therefore never calls the `onValueChange` function with the updated value.

Add optional chaining to the focus function call inside the `setCaretPosition` function.

Screen Shot 2023-12-29 at 2 54 00 PM

- [x] Chrome
- [ ] Chrome (Android)
- [x] Safari (OSX)
- [ ] Safari (iOS)
- [x] Firefox
- [ ] Firefox (Android)

### DIFF
--- a/src/utils.tsx
+++ b/src/utils.tsx
@@ -235,7 +235,7 @@ export function setCaretPosition(el: HTMLInputElement, caretPos: number) {
     }
 
     // fail city, fortunately this never happens (as far as I've tested) :)
-    el.focus();
+    el.focus?.();
     return false;
   }
 }


src/utils.tsx
```diff
@@ -235,7 +235,7 @@ export function setCaretPosition(el: HTMLInputElement, caretPos: number) {
     }
 
     // fail city, fortunately this never happens (as far as I've tested) :)
-    el.focus();
+    el.focus?.();
```
s-yadav, line 238: Q: What is the element type coming here. This function expects the element type to be input. Can you share more details on how the custom input is implemented.

PR #827, opened 2024-02-15
fix: Correct types for customInput property

splatonov92:
Any component containing `value` props can not be passed as `customInput` property. Despite the fact that typing prohibits passing the `customInput` with `value`, `number_format_base` passes all props including `value` to `customInput`. 

- fix customInput type
- remove `@ts-ignore` directive

- [ ] Chrome
- [ ] Chrome (Android)
- [ ] Safari (OSX)
- [ ] Safari (iOS)
- [ ] Firefox
- [ ] Firefox (Android)

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -52,7 +52,7 @@ export type InputAttributes = Omit<
 type NumberFormatProps<Props, BaseType = InputAttributes> = Props &
   Omit<InputAttributes, keyof BaseType> &
   Omit<BaseType, keyof Props | 'ref'> & {
-    customInput?: React.ComponentType<BaseType>;
+    customInput?: React.ComponentType<React.InputHTMLAttributes<HTMLInputElement> & React.RefAttributes<HTMLInputElement>>;
   };
 
 export type OnValueChange = (values: NumberFormatValues, sourceInfo: SourceInfo) => void;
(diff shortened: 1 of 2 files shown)


src/types.ts
```diff
@@ -52,7 +52,7 @@ export type InputAttributes = Omit<
 type NumberFormatProps<Props, BaseType = InputAttributes> = Props &
   Omit<InputAttributes, keyof BaseType> &
   Omit<BaseType, keyof Props | 'ref'> & {
-    customInput?: React.ComponentType<BaseType>;
+    customInput?: React.ComponentType<React.InputHTMLAttributes<HTMLInputElement> & React.RefAttributes<HTMLInputElement>>;
```
s-yadav, line 55: This will force the customInput to be type of Input element. While custom Input can be any component, and the NumberFormatProps inherits from there.

splatonov92: let me try to fix it, may be I need to rewrite a couple of types

src/types.ts
```diff
@@ -46,7 +46,7 @@ export type ChangeMeta = {
 
 export type InputAttributes = Omit<
   React.InputHTMLAttributes<HTMLInputElement>,
-  'defaultValue' | 'value' | 'children'
```
splatonov92, line 49: FMI: why do you need to omit `'defaultValue' | 'value'` from InputAttributes?

PR #836, opened 2024-04-16
- Fix thousandSeparator missing from documentation

ahmad-shoja:
The documentation is missing the thousandSeparator prop in the custom input, causing confusion.


s-yadav:
Thanks. LGTM.

Issue #812, opened 2023-11-21
onValueChange not always fire


alexanderoskin:
#### Describe the issue and the actual behavior
We are trying to implement React Native support like this.

```js
import {useState} from "react";
import {TextInput} from "react-native";
import {NumericFormat} from "react-number-format";

function ControlledNumericFormat({value: propValue, onValueChange}) {
    const [value, setValue] = useState(propValue);

    function handleChange(values) {
        console.log(values);
        onValueChange?.(values.floatValue);
    }

    function handleInputChange(e) {
        setValue(e);
    }

    return (
        <NumericFormat
            value={value}
            onValueChange={handleChange}
            thousandSeparator={","}
            displayType="text"
            renderText={(value) => (
                <TextInput
                    value={value}
                    onChangeText={handleInputChange}
                />
            )}
        />
    );
}
```

All is good, working as expected. But only with numbers longer than three characters. If you type for example 1234 onValueChange is fired. If you type 12 onValueChange is not fire.

#### Provide a CodeSandbox link illustrating the issue
Here is CodeSandbox reproduction with html input https://codesandbox.io/s/controlled-numeric-format-forked-k6q6mf?file=/src/App.js

Try to enter the following numbers
1 - onValueChange is NOT fire
12 - onValueChange is NOT fire
123 - onValueChange is NOT fire
1234 - onValueChange is fire
12345 - onValueChange is fire and so on

Try to remove last character from 12345 number
1234 - onValueChange is fire
123 - onValueChange is fire
12 - onValueChange is NOT fire
1 - onValueChange is NOT fire
clean - onValueChange is NOT fire

hazulu:
I also came across this issue today

rfviolato:
I have the same issue :/

SemBakkumBitvavo:
I also encounter this issue. It's quite unreliable since I perform a set of calculations when either the user changes the value or it's programmatically changed. Would be highly appreciated to look into this?

rfviolato:
[I have created a minimum reproducible example in CodeSandbox](https://codesandbox.io/p/sandbox/numeric-number-format-on-value-change-bug-qrgj2h?file=%2Fsrc%2FApp.tsx%3A56%2C15).

I have added two examples there, one using the `customInput` prop and another using the `renderText` prop. I use the output of the `onValueChanged` to set a local component state to then render it in the UI. The issue is reproducible with both.

[...]

PR #842, opened 2024-05-17
fix: `onValueChange` not called for values with 3 or less characters

rfviolato:
`onValueChange` is not called for values with 3 or less characters.
[The problem is better described in this issue's discussion.](https://github.com/s-yadav/react-number-format/issues/812)

https://codesandbox.io/p/sandbox/numeric-number-format-on-value-change-bug-qrgj2h?file=%2Fsrc%2FApp.tsx

I have removed one of the tests because it doesn't seem to be very meaningful. In the PR's diffs, I have elaborated it in further detail.

This pr aims to fix the issue by simplifying the `useEffect`'s condition which is responsible for ensuring `onValueChange`'s callback is invoked.

https://github.com/s-yadav/react-number-format/issues/812

n/a

n/a

- [x] Chrome
- [ ] Chrome (Android)
- [ ] Safari (OSX)
- [ ] Safari (iOS)
- [ ] Firefox
- [ ] Firefox (Android)

### DIFF
--- a/src/number_format_base.tsx
+++ b/src/number_format_base.tsx
@@ -191,17 +191,14 @@ export default function NumberFormatBase<BaseType = InputAttributes>(
 
   /**
    * if the formatted value is not synced to parent, or if the formatted value is different from last synced value sync it
-   * we also don't need to sync to the parent if no formatting is applied
    * if the formatting props is removed, in which case last formatted value will be different from the numeric string value
    * in such case we need to inform the parent.
    */
   useEffect(() => {
     const { formattedValue: lastFormattedValue, numAsString: lastNumAsString } =
       lastUpdatedValue.current;
-    if (
-      formattedValue !== lastFormattedValue &&
-      (formattedValue !== numAsString || lastFormattedValue !== lastNumAsString)
-    ) {
+
+    if (formattedValue !== lastFormattedValue || numAsString !== lastNumAsString) {
       _onValueChange(getValueObject(formattedValue, numAsString), {
         event: undefined,
         source: SourceType.props,
--- a/test/library/input.spec.jsx
+++ b/test/library/input.spec.jsx
@@ -439,27 +439,6 @@ describe('NumberFormat as input', () => {
     });
   });
 
-  it('should not call onValueChange if no formatting is applied', async () => {
-    const mockOnValueChange = vi.fn();
-
-    const { rerender } = await render(<NumericFormat value="" onValueChange={mockOnValueChange} />);
-
-    expect(mockOnValueChange).not.toHaveBeenCalled();
-
-    rerender(<NumericFormat value={NaN} onValueChange={mockOnValueChange} />);
-    expect(mockOnValueChange).not.toHaveBeenCalled();
-
-    rerender(<NumericFormat value={1234} onValueChange={mockOnValueChange} />);
-    expect(mockOnValueChange).not.toHaveBeenCalled();
-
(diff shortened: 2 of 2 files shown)


src/number_format_base.tsx
```diff
@@ -191,17 +191,14 @@ export default function NumberFormatBase<BaseType = InputAttributes>(
 
   /**
    * if the formatted value is not synced to parent, or if the formatted value is different from last synced value sync it
-   * we also don't need to sync to the parent if no formatting is applied
```
rfviolato, line 194: This behaviour contradicts what is described in the [documentation](https://s-yadav.github.io/react-number-format/docs/props/#onvaluechange-values-sourceinfo--):

>This handler provides access to any values changes in the input field and **is triggered only when a prop changes or the user input changes**

`value` is a prop, so any change should trigger the `onValueChange`. If the original intent of the function was to trigger on formatting changes, it should have been better described and/or be a different prop.

test/library/input.spec.jsx
```diff
@@ -439,27 +439,6 @@ describe('NumberFormat as input', () => {
...
-    expect(mockOnValueChange).not.toHaveBeenCalled();
-
-    rerender(<NumericFormat value={1234} onValueChange={mockOnValueChange} thousandSeparator />);
-    expect(mockOnValueChange.mock.lastCall[0]).toEqual({
-      formattedValue: '1,234',
-      value: '1234',
-      floatValue: 1234,
-    });
-  });
-
```
rfviolato, line 462: This doesn't seem to be a correct test, why should `onValueChange` _not_ be called when the input goes from a `NaN` value to a meaningful value? I have tried to capture this, and further scenarios in the test I have added below, please let me know if this should be adjusted or kept instead of being removed.

rfviolato:
@s-yadav please check whenever you can, this is quite an intrusive bug. The fix seems to work well based on manual testing done both in the repo's example and in the code base I'm using in this library. Also, all unit tests are passing.

rfviolato:
I noticed one case that the fix isn't covering yet. This still needs more work

s-yadav:
Hey @rfviolato , this check was added to avoid double rendering, basically if the parent is controlling the state, and the state is not changed from the rnf, then don't call it to avoid possible double render from the parent setState on onValueChange.

But I agree, the behaviour of onValueChange is inconsistent and makes it unreliable to put some logic, and lib shouldn't try to optimize for things happening from parent.

Let's go with your fix. What's the case is not covered by the fix?

rfviolato:
@s-yadav Thanks for the reply.

Sorry that was a false negative, I've been using a `patch-package` with the fix and had it applied to the wrong file 😅. The fix has been working fine!

>But I agree, the behaviour of onValueChange is inconsistent and makes it unreliable to put some logic, and lib shouldn't try to optimize for things happening from parent. Let's go with your fix. 

Nice! 🚀

Since this issue was affecting a feature my team is developing in the code base we maintain, let me run a few more tests today, and I'll get back to you go give you the thumbs up, then we could proceed with the fix.

rfviolato:
@s-yadav We have been testing the fix and we have also been developing on top of it and it has fixed our issues with `onValueChange`, so the fix seems to be fully working. Shall we proceed?

s-yadav (approved):

s-yadav:
Yes, lets merge this, will make it part of next release.

s-yadav:
@rfviolato this is release on 5.4.0

rfviolato:
Thanks @s-yadav !

PR #847, opened 2024-07-10
react peer dependencies update to ^18.2.0

olga-cm:
didnt test this, but seems logical based on dependencies array


s-yadav:
^18.0.0 is more inclusive to all 18 versions. So it does support 18.2.0 as well.

PR #854, opened 2024-08-30
Fix Number Formatting Order in Customization

mhl98:

Problem: When typing Persian numbers from 1 to 9, there is an issue with how the numbers are displayed. Currently, the numbers appear in an incorrect order or format. For example, if you type a sequence of numbers like ۱۲۳,۴۵۶,۷۸۹  it appears as ۱۲۳,۵۶۸,۹۷۴.

https://github.com/user-attachments/assets/a0c7d9ed-aa50-4bb4-ba00-90292903911e

Expected Behavior: After the fix, when typing Persian numbers from 1 to 9, they should be displayed in the correct order and format, like ۱۲۳,۴۵۶,۷۸۹.

https://github.com/user-attachments/assets/86047fc6-0a31-4cf2-9e7f-c29b84c91810

- [x] Chrome
- [ ] Chrome (Android)
- [ ] Safari (OSX)
- [ ] Safari (iOS)
- [ ] Firefox
- [ ] Firefox (Android)

### DIFF
--- a/documentation/v5/docs/customization.md
+++ b/documentation/v5/docs/customization.md
@@ -186,16 +186,19 @@ Another example for NumericFormat could be support for custom numerals.
 const persianNumeral = ['۰', '۱', '۲', '۳', '۴', '۵', '۶', '۷', '۸', '۹'];
 
 function CustomNumeralNumericFormat(props) {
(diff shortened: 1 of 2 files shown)


s-yadav:
Thanks for the PR. Changes looks good. Merging.

s-yadav (approved):

